PR #5, opened 2026-07-13
fix: pairing code reliability on Windows/Bun

pdfluxo-sys:
## Problem
On Windows/Bun, `requestPairingCode` can hang indefinitely when the underlying
WebSocket is already dead, and a 428 disconnect during pairing isn't always
followed by a clean retry. Also hit .env parsing issues with BOM on Windows-
saved files.

## Fixes
- Parse `.env` robust to BOM/CRLF (Windows-saved files commonly have both)
- Persist the pairing code to a file so `/status` can still show it after a
  process restart, with a fallback read if the in-memory value was lost
- Race `requestPairingCode` against socket disconnect + a hard 20s timeout,
  on both the PHONE_NUMBER and QR paths — reacting to disconnect immediately
  instead of waiting on a timeout that doesn't fire reliably on Bun
- Log pairing errors to `pairing-errors.log` for easier debugging
- Reset `pairingCodeRequested` on a 428 disconnect so the next retry
  re-requests the code instead of getting stuck

No changes to access control, broadcast handling, or cron logic — scoped
entirely to pairing reliability.

Tested on Windows 11 / Bun, repro'd via repeated 428 disconnects during
pairing.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -51,12 +51,15 @@ const MESSAGE_LOG = join(STATE_DIR, 'messages.jsonl')
 const LOCK_FILE = join(STATE_DIR, '.server.lock')
 
 // Load ~/.whatsapp-channel/.env into process.env. Real env wins.
+try { chmodSync(ENV_FILE, 0o600) } catch {}
 try {
-  chmodSync(ENV_FILE, 0o600)
-  for (const line of readFileSync(ENV_FILE, 'utf8').split('\n')) {
+  const envRaw = readFileSync(ENV_FILE, 'utf8')
+  // Strip BOM and normalize CRLF — Notepad/Windows-authored .env files commonly have both.
+  const envText = envRaw.charCodeAt(0) === 0xFEFF ? envRaw.slice(1) : envRaw
+  for (const line of envText.split(/\r?\n/)) {
     const m = line.match(/^(\w+)=(.*)$/)
     if (m && process.env[m[1]] === undefined) {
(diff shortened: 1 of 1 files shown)


nicojones:
Similar to https://github.com/Rich627/whatsapp-claude-plugin/pull/6, both addressing the same issue. So it's probably a real problem, not just me

Rich627:
@codex review

Rich627:
Thanks — you and #6 independently hit the same real problem, which is what confirmed it was worth root-causing. Two pieces of this PR shipped in **v0.11.1** (ae168f6, applied manually since main had moved):

- **`.env` parsing tolerant of BOM/CRLF** — taken nearly verbatim, including splitting the `chmod` into its own try so a chmod failure no longer skips env loading.
- **Racing `requestPairingCode` against socket-close + a 20s timeout** — your diagnosis that a request over a dead socket hangs forever was right; the guard now lives in the centralized request helper from #6.

Two parts were deliberately not taken:

- **Persisting the code to `pairing-code.txt`** — the root cause established in #6 is that a pairing code is only valid on the socket that registered it, so a code from a previous process can never work; surfacing it from disk after a restart would show users a guaranteed-dead code (Codex's review flagged the same staleness issue).
- **The 5s→1.5s pre-request delay** — the shorter window risks racing an unfinished handshake, and with re-registration on reconnect the long delay is no longer costly.

Closing since the surviving pieces are on main. The commit as pushed missed the `Co-authored-by` trailer it should have carried — noting here that the .env hardening and the timeout-race idea are yours. Appreciate the report and the Windows testing.

PR #7, opened 2026-07-15
feat(reply): support @-mentions

nicojones:
## The problem

The `reply` tool cannot tag anyone, and fails silently when asked to.

Writing `@12345` into the message text renders as literal text and notifies nobody. WhatsApp/Baileys only create a real mention when `sendMessage` is passed a **`mentions` array of JIDs**. `server.ts` sends `sock.sendMessage(chat_id, { text })` with no `mentions`, and the `reply` tool had no parameter to supply one — it *reads* `mentionedJid` (for the require-mention gate) but never writes one.

So an agent asked to "tag Lisa so she sees this" produces a message that looks correct in the transcript, reports success, and pings no one. That is the worst shape a bug can take: the caller believes the notification went out.

## The fix

Add an optional `mentions` array to the `reply` tool.

- **Lenient input, canonical output.** Ids are accepted in whatever shape the caller has — bare, `@`-prefixed, LID or phone, or a full JID — and normalised through `lidMap`. The `@lid` form is preferred because group participants are LID-addressed and the in-text convention is `@<lid>`; that is how inbound mentions already arrive, so round-tripping an id from an inbound `<channel>` block works without the caller knowing anything about LIDs.
- **Sensible fallbacks.** Unknown numbers fall back to `<n>@s.whatsapp.net`, duplicates collapse, junk drops.
- **Per-chunk targeting.** A long reply is split into chunks; a mention is attached only to the chunk whose text actually references it, so nobody gets pinged once per chunk. Both text paths (normal chunked send and the document-mode preview) pass them through.

The array and the text must agree — the tool description states this explicitly, since the array alone renders nothing and the text alone notifies nobody.

## Testing

- `node --experimental-strip-types --check server.ts` passes.
- `normalizeMentionJids` verified against a real `lid-map.json`: lid / `@lid` / phone / full JID all collapse to the same `<lid>@lid`; unknown numbers fall back to `@s.whatsapp.net`; duplicates dedupe; junk drops.
- Exercised end-to-end in a live group: a reply carrying `mentions` sends and delivers.
- **Caveat:** I could not programmatically verify the *rendered* mention, as there is no read-back of a sent message's `mentionedJid` — delivery and normalisation are confirmed, the highlight/notification was confirmed only by eye.

Independent of #5 and #6 — this touches only the reply path and the two helpers above, and shares no lines with the pairing work.

### DIFF
--- a/server.ts
+++ b/server.ts
@@ -267,6 +267,35 @@ function resolveToPhone(jid: string): string {
   return lidMap[jidNormalizedUser(jid)] ?? jid
 }
 
+// ─── Outbound mentions ──────────────────────────────────────────────
+// Writing "@12345" in the text is not enough: WhatsApp only renders a mention —
+// and only notifies the person — when the message also carries a `mentions`
+// array of JIDs. Without it the @ is inert text. Accept ids in whatever shape
+// the caller has (bare, @-prefixed, LID or phone, full JID) and normalise.
+function normalizeMentionJids(raw: string[]): string[] {
+  const out: string[] = []
+  for (const entry of raw) {
+    const s = String(entry ?? '').trim().replace(/^@/, '')
+    if (!s) continue
+    if (s.includes('@')) { out.push(jidNormalizedUser(s)); continue }
+    // Group participants are LID-addressed, and the text convention is "@<lid>",
+    // so prefer the LID form whenever we know it.
+    const asLid = `${s}@lid`
+    if (lidMap[asLid]) { out.push(asLid); continue }
+    const lidForPhone = Object.keys(lidMap).find(k => lidMap[k] === `${s}@s.whatsapp.net`)
+    out.push(lidForPhone ?? `${s}@s.whatsapp.net`)
+  }
+  return [...new Set(out)]
+}
+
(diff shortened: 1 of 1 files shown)


nicojones:
Rich — Claude is a genius, and you don't need to merge if you don't want to, but these changes are locally on my fork and work really well

Rich627:
@codex review

Rich627:
Merged into main (b38ba23) and shipped in **v0.12.0** — thanks, this closes a real gap and the implementation was thoughtful: the id normalization (bare/LID/phone/JID) and the chunk-aware attachment (so nobody gets pinged once per chunk) both survived review untouched.

One follow-up landed on top (d84a546), from Codex's review of this PR: in document mode the 200-char preview is the only text message sent, so a mention whose `@id` text fell beyond the preview cut was silently dropped. The preview now carries every requested mention regardless of where the `@id` sits in the text — WhatsApp notifies from the `mentions` array, not the rendered text.

Claude being the one who figured this out from repeated debugging is a nice touch. Appreciated the PR trilogy — between this, #6, and #8 you've materially improved first-run experience for everyone.

PR #9, opened 2026-07-17
fix: fetch WhatsApp Web version at runtime instead of hardcoding (fixes 405 pairing failure)

jawadbokhari:
The hardcoded WhatsApp Web version in `patch-baileys.mjs` goes stale as WhatsApp bumps
the build, and WhatsApp then rejects pairing with `stream:error code 405`
("could not connect" on the phone). This replaces the hardcode with a runtime call to
`fetchLatestBaileysVersion()` (which falls back to the bundled default offline) and
removes the now-redundant version patches. Verified: pairing completes and the device
links after the change.

## Symptoms

```
whatsapp channel: Pairing code: XXXXXXXX
...
whatsapp channel baileys: {"fullErrorNode":{"tag":"stream:error","attrs":{"code":"405"}}}
whatsapp channel: disconnected (reason: 405)
whatsapp channel: reconnecting in ...
```

## Fix

- `server.ts`: call `fetchLatestBaileysVersion()` in `connectWhatsApp()` and pass the
  result into `makeWASocket({ version, ... })`.
- `patch-baileys.mjs`: remove the now-redundant hardcoded version patches (Patch 4).

### DIFF
--- a/patch-baileys.mjs
+++ b/patch-baileys.mjs
@@ -5,7 +5,10 @@
  * 1. passive: true → false  (causes device_removed disconnect)
  * 2. delete lidDbMigrated    (unrecognized field, rejected by WA)
  * 3. remove await on noise.finishInit()  (race condition)
- * 4. update WA Web version (old version rejected with 405)
+ *
+ * NOTE: The WA Web version is intentionally NOT patched here anymore.
+ * server.ts calls fetchLatestBaileysVersion() at runtime, so the version
+ * stays current automatically and never goes stale (was causing 405).
  */
 
 import { readFileSync, writeFileSync, existsSync } from 'fs'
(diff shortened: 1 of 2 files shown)


Rich627:
@codex review

Rich627:
Thanks for the clear diagnosis and the verified repro — you correctly identified the stale hardcoded version as the root cause of the 405 pairing failures.

This ended up being fixed on main in 02928c9 (shipped in v0.10.3) with the same core idea — `fetchLatestBaileysVersion()` — moved to connect time in `server.ts`, plus one hardening step that Codex's review of this PR also flagged: when the fetch fails, `fetchLatestBaileysVersion()` returns the bundled rc.9 default with `isLatest: false`, which is exactly the version WhatsApp rejects. Main's implementation falls back to the last known-good fetched version (or a verified pin), never the bundled default, and memoizes for 6h so reconnect loops add no network dependency.

Closing as superseded — appreciate the report and the verification work.

PR #14, opened 2026-08-19
feat: work under MCP clients that are not Claude Code (conflict mode, portable delivery, terminal access control)

MrTig-afk:
Two commits, both aimed at the same thing: making this plugin usable from an MCP client that is not Claude Code. Reviewable separately, `8fa85e4` then `ea91d40`.

- **`8fa85e4`** - a second server no longer exits silently; it stays up and says who holds the connection. Also three bugs found on the way, one of which was eating real WhatsApp messages.
- **`ea91d40`** - `wait_for_messages` plus an unreplied count on every tool result, so a client with no push channel still finds out about traffic.
- **`7e5af07`** - `scripts/access.ts`, so pairing and allowlisting can be done from a terminal instead of only through a Claude Code skill, plus per-client setup docs.
- **`6cc70ce`** - the watchdog test suite skips itself on Windows rather than failing there, for the same portability reason part 1 fixed in `doctor.test.ts`.

Originally opened as this PR plus #15; consolidated here, since the parts share one motivation and splitting them made the diffs harder to read rather than easier.

## Part 1: a refused server explains itself instead of exiting silently

A second server called `process.exit(2)` from `acquireSingletonLock()`, which runs at module load, ~1500 lines before `mcp.connect()`. The refusal therefore happened **before the MCP handshake existed**, so it only ever reached stderr on a process that was already gone. No MCP client reports why a server died, Claude Code included, so the user just sees a server with no tools and no reason. That is the bug behind every "my WhatsApp tools disappeared" report.

**Change: the loser stays up in conflict mode instead of exiting.**

- `acquireSingletonLock()` returns the holder (`{pid, client}`) instead of exiting. On conflict the server completes the handshake and serves exactly one tool, `whatsapp_unavailable`, plus `instructions` that name the holder. `connectWhatsApp()` is never called.
- **The invariant is unchanged**: one *connection*, not one process. Baileys is untouched in conflict mode, and `releaseSingletonLock()` refuses to delete a lock that is not ours.
- Only `whatsapp_unavailable` is exposed — deliberately not `reply`/`react`. A `reply` that accepts a message and silently fails to send it is worse than no `reply` at all, because the agent reports it as delivered.
- Instructions and tool descriptions are the only channels every MCP client must surface to the model, which is why this works without any client-specific code.

**The lock gains a third line: the `CLAUDE_PID` of the client that started the holder.** That is what lets a refused server, and `doctor`, say *whose* connection this is with a string compare rather than a process query. It is also why this needs no SessionStart hook: only the process that was actually refused ever speaks, so it cannot false-positive on the session's own server — the flaw that pulled the hook out of #13. Both readers destructure earlier lines, so pre-0.15.0 locks stay readable and report "started by a client that does not identify itself, or before 0.15.0".

Why an env var and not process ancestry: under Git Bash on Windows a hook's own `$PPID` is `1`, and the server's parent is the `bun run` wrapper, not the client. `CLAUDE_PID` is present in the spawned environment and costs nothing to compare.

## Three bugs found on the way

1. **`PERMISSION_REPLY_RE` swallowed ordinary messages.** The intercept fired on any `yes|no` + 5 lowercase letters, reacted, and returned early — the text never reached the model. On any client that never sends permission requests, nothing was listening at all, so a genuine WhatsApp message like "yes hello" vanished with only a tick as feedback. Now gated on that request id actually being outstanding, and claiming it consumes it.
2. **`doctor.test.ts` could never pass on Windows.** Its `lstart()` helper called `ps -p <pid> -o lstart=`, the exact probe #13 fixed everywhere else. Same platform branch as the code it tests; doctor tests: 16/17 → 17/17.
3. **Harness check 1b was flaky.** It compared 18-digit FILETIMEs as `[double]`, which exceeds a 53-bit mantissa. Demonstrated on the values that failed: `[double]` → False, `[decimal]` → True. Now `[decimal]`.

Also: `USAGE.md` claimed inbound delivery was broken by client bugs "as of v2.1.83". It works on v2.1.235, verified in a live session. The honest limitation is that it is Claude Code only, and the note now says that.

### How I tested part 1

Windows 11 (10.0.26200) · Bun 1.3.14 · isolated `WHATSAPP_STATE_DIR`. The full harness from #13 was re-run with two checks added and three rewritten for the new contract.

| # | Check | Result |
|---|---|---|
| 1a | duplicate does not connect: stays up, names the live pid, lock still the holder's | PASS |
| 9a | lock line 3 records the starting client's id | PASS |
| 9b | doctor names the owner: own client vs another client | PASS |
| 6 | simultaneous double start, 3 trials → exactly one CONNECTS | PASS |
| 6b | stale lock + simultaneous double start, 5 trials → exactly one CONNECTS | PASS |
| 4a | probe unknown + live pid → conflict mode, no connection (fail closed) | PASS |
| — | the other 16 checks from #13, unchanged | PASS |

Checks 6 and 6b changed assertion, not intent: since the loser no longer exits, "exactly one process survives" became "exactly one CONNECTS", asserted from the winner logging `starting` and the loser logging the conflict. That is the invariant that actually matters.

Additionally, end-to-end against a real running server: a copy of a live lock placed in an isolated state dir with no auth files. The new server read it, entered conflict mode, exposed only `whatsapp_unavailable`, and never touched Baileys. Also confirmed zero non-JSON stdout lines over a real MCP session, since at least one client (Cline) kills a server that emits any.

`tsc --noEmit --strict` clean, `prettier --check` clean. Not exercised: macOS and Linux at runtime (POSIX paths unchanged; `ps -o lstart=` is the same probe as before), and two real Claude Code sessions through the MCP launcher (same code path as 1a).

<details>
<summary>Full harness summary (post-change run)</summary>

#### Summary
| Check | Result |
|---|---|
| 1a duplicate does not connect: stays up in conflict mode naming the live pid, lock still the holder's | PASS |
| 1b lock line 2 == OS start time of the lock holder (to the microsecond) | PASS |
| 5a doctor: live lock reported PASS, no ppid/orphan noise | PASS |
| 5b doctor: stale lock detected | PASS |
| 2a lock released on clean shutdown | PASS |
| 2b dead-pid lock taken over | PASS |
| 2c reused-pid (live unrelated process, wrong start time) taken over | PASS |
| 2d legacy pid-only lock taken over | PASS |
| 2e garbage lock taken over | PASS |
| 2f empty lock taken over | PASS |
| 2g negative-pid lock taken over | PASS |
| 2h injection-shaped lock: server still starts | PASS |
| 2i injection-shaped lock: payload did NOT execute (canary absent) | PASS |
| 6 simultaneous double start: exactly one CONNECTS (3/3 trials) | PASS |
| 9a lock line 3 records the starting client's id | PASS |
| 9b doctor names the owner: own client vs another client | PASS |
| 6b stale lock + simultaneous double start: exactly one CONNECTS (5/5 trials) | PASS |
| 4a probe unknown + live pid â†’ conflict mode, no connection (fail closed) | PASS |
| 4b probe unknown + dead pid â†’ taken over, empty start line, warning logged | PASS |
| 4c doctor: probe unknown â†’ 'could not verify', no rm fix offered | PASS |
| 3 orphan: no false positive for 45 s, exit within ~30 s of parent death, lock released | PASS |
| 7 stdin EOF still shuts down promptly and releases the lock | PASS |

Live bridge PIDs after: 12876, 29648, all pre-existing PIDs still alive: True
Stray test processes after cleanup: 0

VERDICT: SHIP
Done 12:33:06.988.

</details>

## Part 2: clients that cannot be pushed to can still find out

Inbound messages reach a session through `notifications/claude/channel`, a Claude Code extension. I went looking for the standard equivalent and there isn't one:

- Every standard server-to-client message is either request-scoped (`notifications/progress`, and in the current spec `notifications/message` too) or subscription-gated (`resources/updated`, which needs the client to have called `resources/subscribe` first). None is defined as putting text in front of the model; resources are explicitly "application-driven, with host applications determining how to incorporate context".
- Unknown notification methods are dropped with no error and no delivery signal. JSON-RPC forbids responding to a notification, and the reference SDK's `_onnotification` returns early when no handler is registered, so a server cannot even tell that delivery failed.
- The 2026-07-28 spec deprecates sampling and logging and states the server **MUST NOT** send notification types the client has not explicitly requested.
- `sampling/createMessage` cannot help even in principle: its result returns to the server, not into the user's conversation.

So on Codex CLI, Gemini CLI or Cursor, a WhatsApp message today arrives, passes the access gate, gets logged, and nobody ever mentions it. This adds the two mechanisms that need no capability negotiation and therefore work on any client.

**`wait_for_messages`** returns unreplied messages, waiting up to 40 seconds for the first to land. 40s and not longer because the reference SDK's `DEFAULT_REQUEST_TIMEOUT_MSEC` is 60s and `resetTimeoutOnProgress` defaults to false, so an over-long wait is cancelled client-side rather than answered. Measured 40.1s, leaving 19.9s of margin. An empty return is normal, not an error, and the description says so.

**Every tool result now carries the unreplied count.** A client that cannot be pushed to still learns there is traffic on its next tool call, whatever that call was.

Claude Code is unaffected: the channel notification still fires and remains the fast path there.

Kept small deliberately: the `unreplied` formatter is now shared instead of duplicated, and the tool handler is wrapped rather than rewritten, so the switch body is untouched in the diff.

### How I tested part 2

Over a real MCP session against the server (initialize, `tools/list`, `tools/call`), with inbound messages simulated by writing to `messages.jsonl`, the same file the real inbound path writes.

| Behaviour | Result |
|---|---|
| nothing pending → waits rather than returning empty | PASS (still waiting at 8s) |
| message lands mid-wait → returns early | PASS (~1s after it landed, not at the cap) |
| already pending → returns immediately | PASS |
| count appended to an unrelated tool's result (`status`) | PASS |
| timeout path returns rather than hanging | PASS (40.1s, 19.9s margin under a 60s client timeout) |
| non-JSON stdout lines | 0 |

That last row is deliberate: at least one client (Cline) treats any non-JSON line on stdout as a protocol failure and kills the child process, so it is worth asserting rather than assuming.

`tsc --noEmit --strict` clean, `prettier --check` clean, the #14 lock harness still 22/22, doctor tests 17/17.

Known ceiling, marked with a comment in the code: the wait re-reads the message log every 2 seconds rather than being woken in-process. That is simpler, works no matter which code path wrote the line, and costs at most 2s of latency in a chat bridge. An in-process wake is the upgrade if that ever matters. I had the wake registry written and removed it: it duplicated the re-check, and it had silently failed a probe in a way the re-check caught.

## Part 3: setup without Claude Code

Pairing, allowlisting, group setup and delivery config lived only in `skills/access/SKILL.md`, which is a document Claude Code executes. Under Codex CLI, Gemini CLI or Cursor there was no way to approve a contact except editing `access.json` by hand, which made the plugin effectively unusable there however well the tools worked. Of the three gaps this PR closes, that was the biggest.

`scripts/access.ts` covers the same surface as the skill: `status`, `pair`, `deny`, `allow`, `remove`, `policy`, `group add/rm`, `set`. It writes the same `access.json` and the same `approved/<senderId>` handoff the server polls, so the two are interchangeable. The skill stays the friendlier path, since it can interview the user and write a tailored group `config.md`; the CLI seeds a default one and tells you where to edit it. Both are cross-referenced so neither is a dead end.

Security properties kept deliberately, and covered by tests:

- Approving always requires the specific code, even when only one pairing is pending. Anyone can create a pending entry just by messaging the account, so "approve the pending one" is exactly what an injected request looks like.
- The pairing flow re-locks `dmPolicy` to `allowlist` once nothing else is waiting, so an open pairing window is never left behind.
- It is a terminal command and deliberately **not** an MCP tool, so nothing arriving over WhatsApp can reach it.
- A corrupt `access.json` is reported, never silently replaced.

README gains per-client registration for Codex CLI, Gemini CLI and Cursor, with absolute paths since `${CLAUDE_PLUGIN_ROOT}` is substituted by Claude Code only, and a plain statement that inbound messages are poll-based outside Claude Code rather than leaving people to discover that.

### How I tested part 3

12 new tests in `scripts/access.test.ts`, all passing, covering the paths worth being sure about rather than every branch: a wrong code and an expired code each approve nobody, a valid one allows exactly the sender and writes the server handoff, the policy stays open while another pairing is still waiting, a hand-edited group `config.md` survives re-adding the group, typed `set` values are coerced or refused, and a corrupt file is left byte-for-byte intact.

Also verified end to end rather than in isolation: after driving the CLI through a full setup, `doctor` reports `[PASS] access-config: dmPolicy: allowlist, 1 allowed contact(s), 1 group(s)` and the server starts against that state dir and parses it with no access-related errors. A config only the CLI understood would be worthless.

Repo test suite on Windows: **29 pass, 13 skip, 0 fail** (POSIX runs all 42).

Getting there meant fixing two pre-existing Windows-portability bugs in the test harness, both the same class as the `ps -o lstart=` bug this repo fixed in #13, and neither caused by this PR:

- `doctor.test.ts` computed a process start time with `ps -p <pid> -o lstart=`, so its live-lock test could never pass on Windows. Fixed in part 1 with the same platform branch the code under test uses: 16/17 to 17/17.
- `install-watchdog.test.ts` built `PATH` with a hardcoded `:` separator (`;` on Windows) and stubbed `crontab` as an extensionless shell script, which Windows cannot spawn. Since `install-watchdog` drives `crontab` and `watchdog.sh` calls `launchctl` and `tmux`, there is nothing real to test on Windows, so that suite now skips with a stated reason and uses `path.delimiter` where it does run. Worth flagging the tradeoff: 5 tests that passed incidentally on Windows are now skipped too, on paths that never reached `crontab`.

Note this is local signal only, since no workflow runs `bun test` today. Happy to add a test job if you want the suite gating CI; that felt like your call rather than mine, so I left it out.

## Checklist

- [x] **Version bumped in BOTH files to the same new version** (minor = feature): `0.14.1` -> `0.15.0`
  - [x] `.claude-plugin/plugin.json` -> `version`
  - [x] `.claude-plugin/marketplace.json` -> `plugins[0].version`
- [x] Touches the **singleton lock** (part 1) - invariant described above, and the harness re-run is the gate.
- [x] Lock file format extended additively; old locks still parse.
- [x] New CLI is not exposed as an MCP tool; access control stays reachable only from a terminal.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "whatsapp-claude-channel",
       "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.14.1",
+      "version": "0.15.0",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "whatsapp-claude-channel",
   "displayName": "WhatsApp Channel for Claude Code",
   "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-claude-channel:access.",
-  "version": "0.14.1",
+  "version": "0.15.0",
   "author": {
     "name": "Rich627"
   },
--- a/README.md
+++ b/README.md
@@ -31,13 +31,88 @@ Inside the session, set your number and pair:
 
 A pairing code is printed on first launch. On your phone: WhatsApp → Settings → Linked Devices → Link a Device → **Link with phone number instead** → enter the code. No WhatsApp Business API, Meta developer account, or API key is involved — it links to your regular account.
 
+## Other MCP clients (Codex CLI, Gemini CLI, Cursor)
+
+The server is a plain stdio MCP server, so any MCP client can run it. Two things are Claude Code specific and worth knowing before you start:
+
+- **Inbound messages are not pushed.** Waking a session on an incoming message uses `notifications/claude/channel`, a Claude Code extension. MCP has no standard equivalent that reaches the model, and other clients drop unknown notifications silently. Elsewhere the plugin is poll-based: call `wait_for_messages` (waits up to 40s for the next message) or `catch_up` / `unreplied`. Every tool result also carries a count of unreplied messages, so a client finds out there is traffic on its next call whatever that call was.
+- **Setup is done from a terminal, not a slash command.** `/whatsapp-claude-channel:access` and friends are Claude Code skills. Use `bun scripts/access.ts` instead (see [Access control from a terminal](#access-control-from-a-terminal)).
+
+Register the server with an absolute path — `${CLAUDE_PLUGIN_ROOT}` is substituted by Claude Code only:
+
+**Codex CLI** (`~/.codex/config.toml`)
+
+```toml
+[mcp_servers.whatsapp]
+command = "bun"
+args = ["run", "--cwd", "/absolute/path/to/whatsapp-claude-channel", "start"]
+startup_timeout_sec = 30   # default 10 is tight for a first Baileys connect
+tool_timeout_sec = 120     # default 60; wait_for_messages parks for up to 40s
+```
+
+**Gemini CLI** (`~/.gemini/settings.json`)
+
+```json
+{
+  "mcpServers": {
+    "whatsapp": {
+      "command": "bun",
+      "args": [
+        "run",
+        "--cwd",
+        "/absolute/path/to/whatsapp-claude-channel",
+        "start"
+      ],
+      "timeout": 600000
+    }
+  }
+}
+```
+
+**Cursor** (`~/.cursor/mcp.json` for all projects, `.cursor/mcp.json` for one)
+
+```json
+{
+  "mcpServers": {
+    "whatsapp": {
+      "type": "stdio",
+      "command": "bun",
+      "args": [
+        "run",
+        "--cwd",
+        "/absolute/path/to/whatsapp-claude-channel",
+        "start"
+      ]
+    }
+  }
+}
+```
+
+Only one client at a time can hold the WhatsApp connection: WhatsApp allows one linked-device session per account, and two servers would kick each other off. A second server does not fail silently — it stays up and serves a single `whatsapp_unavailable` tool naming the process that holds the connection.
+
+## Access control from a terminal
+
+Everything the access skill does, without Claude Code:
+
+```sh
+bun scripts/access.ts status                 # policy, allowlist, pending codes, groups
+bun scripts/access.ts policy pairing         # open the door
+bun scripts/access.ts pair <code>            # approve someone who messaged you
+bun scripts/access.ts allow <jid>            # add directly
+bun scripts/access.ts remove <jid>
+bun scripts/access.ts group add <groupJid> [--mention] [--allow jid1,jid2]
+bun scripts/access.ts set replyToMode first  # ackReaction, textChunkLimit, chunkMode, mentionPatterns
+```
+
+Approving always needs the specific code, even when only one pairing is waiting: anyone can create a pending entry just by messaging the account, so "approve the pending one" is exactly what a prompt-injected request looks like. For the same reason this is a terminal command and deliberately **not** an MCP tool, so nothing arriving over WhatsApp can reach it.
+
 ## Features
 
 - **Bidirectional messaging.** Send and receive from the session; long replies are chunked to WhatsApp's limits or sent as a document attachment past a configurable threshold.
 - **@-mentions.** `reply` can tag people so they actually get notified — ids are accepted as phone, LID, or full JID, and mentions attach only to the chunk that names them.
 - **Full media support.** Photos, voice notes, video, documents, and stickers, in both directions.
 - **Voice transcription.** Incoming voice notes are transcribed locally via mlx-whisper (see [setup](#voice-transcription-optional)); without the script they arrive as plain attachments.
-- **Access control.** Pairing codes, allowlists, and per-group policies gate every inbound message — strangers never reach your session. Managed via `/whatsapp-claude-channel:access`.
+- **Access control.** Pairing codes, allowlists, and per-group policies gate every inbound message — strangers never reach your session. Managed via `/whatsapp-claude-channel:access` in Claude Code, or `bun scripts/access.ts` anywhere.
 - **Per-group personalities.** Each group gets its own `config.md` with a custom personality and conversation memory.
 - **Permission relay.** Approve or deny Claude's tool requests from WhatsApp with an emoji reaction (👍 / 👎).
 - **Cron tasks.** A `## Cron Jobs` section in a group's `config.md` schedules recurring server-side tasks.
--- a/USAGE.md
+++ b/USAGE.md
@@ -207,7 +207,7 @@ pkill -f "whatsapp.*server"
 
 ## Known limitations
 
-**Inbound message delivery may not work.** Claude Code's channel notification system (`notifications/claude/channel`) has a confirmed client-side bug where inbound messages sent by the MCP server are silently dropped and never appear in the conversation. This affects all channel plugins (WhatsApp, Telegram, etc.) and is tracked across multiple issues ([#37933](https://github.com/anthropics/claude-code/issues/37933), [#36477](https://github.com/anthropics/claude-code/issues/36477), [#37633](https://github.com/anthropics/claude-code/issues/37633)). The server-side implementation is correct — the fix must come from the Claude Code client. No reliable workaround exists as of v2.1.83.
+**Inbound message delivery is Claude Code only.** Messages are pushed into a session with `notifications/claude/channel`, which is a Claude Code extension, not part of MCP. Other MCP clients (Codex CLI, Gemini CLI, Cursor) drop unknown notifications silently, so there the plugin is poll-only: call `catch_up` or `unreplied` to see what arrived. Delivery inside Claude Code was broken by client bugs ([#37933](https://github.com/anthropics/claude-code/issues/37933), [#36477](https://github.com/anthropics/claude-code/issues/36477), [#37633](https://github.com/anthropics/claude-code/issues/37633)) and worked again when last checked on v2.1.235; if messages stop appearing, check those issues before suspecting this plugin.
 
 ## Resetting auth
 
--- a/USAGE.md
+++ b/USAGE.md
@@ -207,7 +207,7 @@ pkill -f "whatsapp.*server"
 
 ## Known limitations
 
-**Inbound message delivery is Claude Code only.** Messages are pushed into a session with `notifications/claude/channel`, which is a Claude Code extension, not part of MCP. Other MCP clients (Codex CLI, Gemini CLI, Cursor) drop unknown notifications silently, so there the plugin is poll-only: call `catch_up` or `unreplied` to see what arrived. Delivery inside Claude Code was broken by client bugs ([#37933](https://github.com/anthropics/claude-code/issues/37933), [#36477](https://github.com/anthropics/claude-code/issues/36477), [#37633](https://github.com/anthropics/claude-code/issues/37633)) and worked again when last checked on v2.1.235; if messages stop appearing, check those issues before suspecting this plugin.
+**Inbound message delivery is Claude Code only.** Messages are pushed into a session with `notifications/claude/channel`, which is a Claude Code extension, not part of MCP. Other MCP clients (Codex CLI, Gemini CLI, Cursor) drop unknown notifications silently, so there the plugin is poll-only: call `wait_for_messages`, `catch_up` or `unreplied` to see what arrived. Delivery inside Claude Code was broken by client bugs ([#37933](https://github.com/anthropics/claude-code/issues/37933), [#36477](https://github.com/anthropics/claude-code/issues/36477), [#37633](https://github.com/anthropics/claude-code/issues/37633)) and worked again when last checked on v2.1.235; if messages stop appearing, check those issues before suspecting this plugin.
 
 ## Resetting auth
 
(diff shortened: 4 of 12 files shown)


MrTig-afk:
Follow-up is up as #15 (portable inbound delivery for clients that cannot be pushed to). It stacks on this one, so its diff currently shows this commit too; I will rebase it once this lands. No rush on it, #14 is the one that unblocks the other.

Rich627:
Thanks for this — the writeup is unusually careful and the reasoning behind each decision is easy to follow. I reviewed all four commits and ran the suite locally (42 pass / 0 fail; `tsc --strict` clean on `server.ts` and `scripts/access.ts` — the two `Bun` global errors in `doctor.ts` are pre-existing, this repo has no `@types/bun`).

One blocker, then some smaller things.

## Blocker: a conflict-mode server deletes the pairing handoff files

Part 1 changes a duplicate server from `process.exit(2)` to staying up. But `server.ts` registers several module-level `setInterval`s, and they all keep running in conflict mode. `checkApprovals` (server.ts:669, every 5s) does:

```js
if (!sock) { rmSync(file, { force: true }); continue; }
```

In conflict mode `connectWhatsApp()` is never called, so `sock` is permanently `null`, and the server deletes every `approved/<senderId>` file within 5 seconds — the handoff that tells the *connected* server to send the pairing confirmation.

Reproduced on an isolated `WHATSAPP_STATE_DIR` with the lock held by a live pid:

| branch | result |
| --- | --- |
| `main` (85eda73) | exits 2; `approved/886900000000@s.whatsapp.net` still there |
| this PR (6cc70ce) | enters conflict mode; after 14s `approved/` is empty |

This lands squarely on the flow Part 3 adds: `bun scripts/access.ts pair <code>` writes the handoff, and the same session's conflict-mode server eats it before the real server can act on it, so the newly paired contact never gets "Paired! Say hi to Claude." It is also a race between the two servers, so it fails intermittently rather than consistently. The contact is still allowlisted (`access.json` is the source of truth), so the failure is silent — which is the exact class of bug this PR set out to remove.

The fix is small:

```js
if (!STATIC && !CONFLICT) setInterval(checkApprovals, 5000).unref();
```

`pruneMessageLog` (server.ts:1067) wants the same guard. It does a full read-then-rewrite of the message log every hour, so a conflict-mode server is now a second writer to that file and can drop messages the connected server appended between its read and its write. I did not reproduce this one (it needs an hour of uptime) — flagging it as reasoning, not as an observed failure. The cron engine is fine, it already returns early on `!sock`.

The general point: "stay up instead of exiting" removes the implicit guarantee that only one process ever touches the state dir, and the PR only audits that against Baileys. Worth walking the module-level side effects as a set.

## Smaller things

**`scripts/access.ts` — the comment above `save()` describes behaviour the code does not have.**

```js
// Always load immediately before saving: the server may have added a pending
// entry between our read and write, and clobbering it loses a pairing request.
function save(access: Access): void {
```

`save()` writes the whole object with no re-read, and callers do `load()` → mutate → `save()` with a real gap in between, so `pair()` can drop a pending entry the server added meanwhile. Either implement the read-modify-write or drop the comment — right now it claims a protection that is not there.

**`group add` argument parsing.** `requireArg(args[1], "group JID")` takes whatever is in position 1, so `group add --mention <jid>` writes a group keyed `--mention` into `access.json` and creates `groups/--mention/`. A `startsWith("--")` check would cover it.

**The unreplied-count wrapper excludes `unreplied` but not `wait_for_messages`.** That tool has just returned the pending messages, and the appended line then tells the model to "call unreplied or wait_for_messages", pointing it back at the same set.

**The 10-minute TTL on `permissionMessageMap` now gates text approvals.** `claimPermission` only succeeds while the entry is live, so a "yes xxxxx" that arrives after 10 minutes is no longer intercepted and reaches the model as ordinary chat. The emoji path already had this ceiling so it is an alignment rather than a new regression, but Claude Code waits on a permission request indefinitely, so it is worth confirming the tradeoff is intentional.

## What I think is right

- The `PERMISSION_REPLY_RE` bug is real and worth the PR on its own — `^(y|yes|n|no)\s+[a-km-z]{5}$` is reachable by ordinary chat, and the old code swallowed the message with only a reaction as feedback. Gating on an outstanding request id is the correct shape of fix.
- Not exposing `reply`/`react` in conflict mode is the right call for the reason you give: accepting a message, failing to send it, and having the agent report success is worse than having no tool.
- `releaseSingletonLock()` returning early on `CONFLICT` keeps "never delete a lock that is not ours".
- Appending the client id as a third line, with both readers destructuring the earlier lines, keeps pre-0.15.0 locks readable. Good compatibility instinct.
- The `USAGE.md` correction is a real improvement — the old text asserted inbound delivery was broken when the actual limitation is that push is Claude Code only.

## Version

The version bump to 0.15.0 collides with another branch of mine that also claims 0.15.0. I will renumber mine — no action needed here, just flagging it so it does not get resolved the other way by accident.

MrTig-afk:
Blocker confirmed and fixed in f412cf8: `checkApprovals` and `pruneMessageLog` now only register when `!CONFLICT`. I walked the rest of the module-level side effects as you suggested: the cron tick already returns on `!sock`, the `sentMessages` sweep is in-memory, the parent watchdog's `shutdown()` goes through `releaseSingletonLock()` which already returns on `CONFLICT`, and the tool handlers are replaced by the conflict ones — so those two were the only state-dir writers. Added `scripts/server-conflict.test.ts`, which spawns the real server against a planted live lock (own pid + start time, same probe) and asserts the handoff file survives; it fails on 6cc70ce and passes now.

Smaller things:
- `save()` comment rewritten to describe the actual behaviour (small load→save window, accepted, the sender just messages again) and the write made atomic like server.ts.
- `group add` now goes through `node:util` `parseArgs` in strict mode, so a flag in the JID slot, an unknown flag, or a valueless `--allow` is an error rather than a mis-keyed group (test added; `--mention` before the JID also just works now).
- `wait_for_messages` excluded from the unreplied suffix.
- TTL: Claude Code waits indefinitely, so I dropped the 10-minute expiry entirely rather than keep the ceiling; entries are freed when claimed. Happy to restore a bound if you'd rather have one.

Version: noted, leaving 0.15.0 here unless you want it renumbered.

Suite: 44 tests (31 pass / 13 skip on Windows), `tsc --strict` clean on server.ts and access.ts.

Rich627:
LGTM — verified `f412cf8` locally: my original repro now leaves the handoff file intact with the lock still the holder's, the full suite is 44/44 here, and I confirmed `server-conflict.test.ts` is a real regression test by swapping in the pre-fix `server.ts` and watching it fail. The side-effect audit matches what I found independently. Thank you for the contribution — the writeups on this and #13 set a high bar.

One heads-up: while re-reading `checkApprovals` I noticed a pre-existing bug that is not yours and not blocking — the `if (!sock) rmSync(file)` branch also fires on a *connected* server during the reconnect window (`sock` is nulled on `connection === "close"`), silently eating pairing handoffs. I'm fixing that on `main` directly, so you may see a tiny merge nudge in that function.

PR #15, opened 2026-08-19
feat(tools): let clients that cannot be pushed to find out anyway

MrTig-afk:
> **Stacked on #14.** GitHub cannot target a branch on my fork, so this PR shows both commits. Only the second one, `ea91d40`, is new here; `8fa85e4` is #14 under review. I will rebase onto main once #14 lands, and this can wait until then.

## What & why

Inbound messages reach a session through `notifications/claude/channel`, a Claude Code extension. I went looking for the standard equivalent and there isn't one:

- Every standard server-to-client message is either request-scoped (`notifications/progress`, and in the current spec `notifications/message` too) or subscription-gated (`resources/updated`, which needs the client to have called `resources/subscribe` first). None is defined as putting text in front of the model; resources are explicitly "application-driven, with host applications determining how to incorporate context".
- Unknown notification methods are dropped with no error and no delivery signal. JSON-RPC forbids responding to a notification, and the reference SDK's `_onnotification` returns early when no handler is registered, so a server cannot even tell that delivery failed.
- The 2026-07-28 spec deprecates sampling and logging and states the server **MUST NOT** send notification types the client has not explicitly requested.
- `sampling/createMessage` cannot help even in principle: its result returns to the server, not into the user's conversation.

So on Codex CLI, Gemini CLI or Cursor, a WhatsApp message today arrives, passes the access gate, gets logged, and nobody ever mentions it. This adds the two mechanisms that need no capability negotiation and therefore work on any client.

**`wait_for_messages`** returns unreplied messages, waiting up to 40 seconds for the first to land. 40s and not longer because the reference SDK's `DEFAULT_REQUEST_TIMEOUT_MSEC` is 60s and `resetTimeoutOnProgress` defaults to false, so an over-long wait is cancelled client-side rather than answered. Measured 40.1s, leaving 19.9s of margin. An empty return is normal, not an error, and the description says so.

**Every tool result now carries the unreplied count.** A client that cannot be pushed to still learns there is traffic on its next tool call, whatever that call was.

Claude Code is unaffected: the channel notification still fires and remains the fast path there.

Kept small deliberately: the `unreplied` formatter is now shared instead of duplicated, and the tool handler is wrapped rather than rewritten, so the switch body is untouched in the diff.

## How I tested it

Over a real MCP session against the server (initialize, `tools/list`, `tools/call`), with inbound messages simulated by writing to `messages.jsonl`, the same file the real inbound path writes.

| Behaviour | Result |
|---|---|
| nothing pending → waits rather than returning empty | PASS (still waiting at 8s) |
| message lands mid-wait → returns early | PASS (~1s after it landed, not at the cap) |
| already pending → returns immediately | PASS |
| count appended to an unrelated tool's result (`status`) | PASS |
| timeout path returns rather than hanging | PASS (40.1s, 19.9s margin under a 60s client timeout) |
| non-JSON stdout lines | 0 |

That last row is deliberate: at least one client (Cline) treats any non-JSON line on stdout as a protocol failure and kills the child process, so it is worth asserting rather than assuming.

`tsc --noEmit --strict` clean, `prettier --check` clean, the #14 lock harness still 22/22, doctor tests 17/17.

Known ceiling, marked with a comment in the code: the wait re-reads the message log every 2 seconds rather than being woken in-process. That is simpler, works no matter which code path wrote the line, and costs at most 2s of latency in a chat bridge. An in-process wake is the upgrade if that ever matters. I had the wake registry written and removed it: it duplicated the re-check, and it had silently failed a probe in a way the re-check caught.

## Checklist

- [ ] **Version bump** deliberately not included: this stacks on #14, which already moves `0.14.1` to `0.15.0`, and a second bump would collide on rebase. Happy to add one (`0.16.0`) if you would rather each PR carry its own.
- [x] Does not touch the singleton lock.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "whatsapp-claude-channel",
       "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.14.1",
+      "version": "0.15.0",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "whatsapp-claude-channel",
   "displayName": "WhatsApp Channel for Claude Code",
   "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-claude-channel:access.",
-  "version": "0.14.1",
+  "version": "0.15.0",
   "author": {
     "name": "Rich627"
   },
--- a/USAGE.md
+++ b/USAGE.md
@@ -207,7 +207,7 @@ pkill -f "whatsapp.*server"
 
 ## Known limitations
 
-**Inbound message delivery may not work.** Claude Code's channel notification system (`notifications/claude/channel`) has a confirmed client-side bug where inbound messages sent by the MCP server are silently dropped and never appear in the conversation. This affects all channel plugins (WhatsApp, Telegram, etc.) and is tracked across multiple issues ([#37933](https://github.com/anthropics/claude-code/issues/37933), [#36477](https://github.com/anthropics/claude-code/issues/36477), [#37633](https://github.com/anthropics/claude-code/issues/37633)). The server-side implementation is correct — the fix must come from the Claude Code client. No reliable workaround exists as of v2.1.83.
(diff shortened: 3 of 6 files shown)


MrTig-afk:
Folding this into #14 instead. The split made the diff harder to read rather than easier: GitHub cannot target a branch on my fork, so this PR showed both commits anyway, and the two halves share one motivation (running under a client that is not Claude Code). #14 now carries both commits, still reviewable one at a time, and there is only one thread to follow.

Also resolves the version question I raised here: one PR, one bump, 0.14.1 -> 0.15.0.

PR #16, opened 2026-08-21
fix(mentions): match the id the caller typed, not the resolved JID

MrTig-afk:
## What this fixes

The bug you'd flagged: four bare-number mentions in one message rendered inconsistently, one of them not at all.

`normalizeMentionJids` silently swaps a bare phone number for a cached LID mapping when one exists (a group-participant addressing preference). But `mentionsForChunk` was matching the message text against that *resolved* JID's local part, not the id the caller actually typed and was told to echo in the text. Whenever a mapping happened to be cached for one of the mentioned people, their id no longer matched, and they silently dropped out of the mentions array. No error, just an inert `@number` in the sent message.

Fix: the match key is now always the caller's original input (its own local part, stripped of any domain), never the resolved JID. Extracted `normalizeMentionJids`/`mentionsForChunk` to `scripts/mentions.ts` so they're unit-testable without pulling in `server.ts`'s connect-on-import side effects.

Two review passes (code-review, run twice) turned up three more edge cases in the same matching logic, all fixed and covered by tests:
- a full JID passed as the mention id matched against the whole string instead of its local part
- deduping by resolved JID silently dropped a second valid spelling of the same person
- a doubled leading `@` produced an empty match key, which built a regex that matched almost any `@` in the text

## Testing

41 pass / 13 skip (Windows, pre-existing/unrelated) / 0 fail. New regression tests reproduce the reported bug directly (fail on the old matching logic, pass on the new one).

## Scope

Only touches mention matching. No other behavior changes.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "whatsapp-claude-channel",
       "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.15.2",
+      "version": "0.15.3",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "whatsapp-claude-channel",
   "displayName": "WhatsApp Channel for Claude Code",
   "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-claude-channel:access.",
-  "version": "0.15.2",
+  "version": "0.15.3",
   "author": {
     "name": "Rich627"
   },
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,3 +35,12 @@ jobs:
           fetch-depth: 0
       - name: Trunk Check
         uses: trunk-io/trunk-action@04ba50e7658c81db7356da96657e6e77f220bfa3 # v1.3.1
+
+  test:
+    name: Test
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@34e114876b0b11c390a56381ad16ebd13914f8d5 # v4.3.1
+      - uses: oven-sh/setup-bun@0c5077e51419868618aeaa5fe8019c62421857d6 # v2.2.0
+      - run: bun install
+      - run: bun test
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -6,13 +6,22 @@ small and moves fast, so a few conventions keep pull requests easy to review and
 ## Project at a glance
 
 - **Runtime:** [Bun](https://bun.sh). TypeScript runs directly — there is **no build
-  step** and **no test suite**.
+  step**.
 - **Dependencies:** only two — `@modelcontextprotocol/sdk` and
   `@whiskeysockets/baileys`. Baileys is pinned to a release candidate and patched by
   `patch-baileys.mjs` (runs automatically on `postinstall`). **Please do not add new
   runtime dependencies without opening an issue first** — a slim dependency tree is a
   deliberate goal.
-- **The whole MCP server is one file:** `server.ts`.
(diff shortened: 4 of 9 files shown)


MrTig-afk:
Small one this time. The mention-rendering fix from the bug report you'd flagged: a LID-cached number silently drops out of the mentions array when the message text still uses the plain number. Tests included, regression covered.

Separately, no action needed here: I'm poking at a privacy-focused pass on how mentions and group access work in this plugin, nothing concrete yet, still shaping it. Will open something separate once there's actually a diff worth looking at.

Rich627 (changes requested):
Thanks — the core fix is right, and the bare-number path is well covered by the regression tests. Matching on the id the caller actually typed instead of the resolved LID is the correct contract. Two things I'd like fixed before merging, plus a couple of smaller asks.

**Blocking:**

1. **Device-suffixed full JIDs regress** (`scripts/mentions.ts` line 47). The match key is `s.split("@")[0]` on the raw input, so `61434505973:12@s.whatsapp.net` — which is exactly the shape `reply_to_sender` surfaces verbatim from `contextInfo.participant` — keeps `:12` in the key. The regex then looks for `@61434505973:12` while the text says `@61434505973`, and the mention silently never attaches. The pre-PR code keyed off the normalized jid's local part and handled this. Suggest deriving the key from `jidDecode(s)?.user` (Baileys already exports it) for full-JID inputs. Related: the tests inject a hand-rolled `jidNormalizedUser` stand-in that diverges from the real Baileys export (the real one doesn't lowercase, and also strips `_agent` suffixes) and there's no device-suffix case — please import the real `jidNormalizedUser` in the tests and add one.

2. **Version bump.** Both `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` (`plugins[0].version`) need a patch bump — CI only checks that the two files match each other, so this merges green as-is, but `plugin update` would silently no-op and the fix never reaches users.

**Also in this PR, please:**

3. `scripts/` has a documented contract in CLAUDE.md: reference scripts users copy out of the repo, where repo edits don't affect running deployments. A module `server.ts` imports at runtime inverts that — please move it to something like `lib/`, and update CLAUDE.md's "entire MCP server in one file" / "No test suite" claims to match reality.

4. CI never runs the new tests — the extraction's whole justification is testability, but nothing executes `bun test`. A small test job in `ci.yml` would keep the `MentionPair[]` contract and these regressions actually enforced.

**Non-blocking, noting for later (happy to take these as a separate issue):** default length-mode chunking can bisect `@<id>` at the 4096-char cut, after which the mention matches no chunk — same bug class as this PR, pre-existing; and keeping duplicate jids in `MentionPair[]` pushes deduping onto every consumer and allows a narrow cross-chunk double-ping — grouping as `{ jid, inputs: string[] }` at construction would close both.

MrTig-afk:
Fixed both blocking items and the two "also please" asks:

1. **Device-suffixed full JIDs** - the match key for a full-JID input now comes from `jidDecode(s)?.user` instead of `s.split("@")[0]`, so a device suffix (`:12`) no longer survives into the regex match. Added a regression test with a device-suffixed input. Also swapped the tests' hand-rolled `jidNormalizedUser` stand-in for the real Baileys export and imported `jidDecode` from there too.
2. **Version bump** - `plugin.json` and `marketplace.json`'s `plugins[0].version` both bumped 0.15.2 -> 0.15.3 (top-level marketplace version left alone, per CONTRIBUTING.md's warning).
3. **`scripts/` -> `lib/`** - moved `mentions.ts`/`mentions.test.ts` since `server.ts` imports it at runtime; updated CONTRIBUTING.md's "no test suite" / "whole server is one file" claims to describe the actual `lib/` vs `scripts/` split and the existing test suite.
4. **CI test job** - added a `test` job to `ci.yml` (`bun install` + `bun test`) so the suite actually runs on every PR, not just locally.

Left the two non-blocking items (chunk-boundary mention splitting, `MentionPair[]` dedup) as you suggested - happy to take those as a separate issue.

Rich627 (approved):
All four items verified against 48f367a — device-suffix key via jidDecode with a regression test on the real Baileys exports, both versions bumped to 0.15.3, module moved to lib/ with the contract documented, and the test job actually running in CI. Thanks for the quick turnaround. Merging; feel free to rebase #17 on top.

PR #17, opened 2026-08-21
Privacy-focused names, mentions and group access

MrTig-afk:
## Summary

A privacy-focused rework of how the plugin handles names, mentions and access, for both groups and DMs.

- Contact names sync from the account's own Baileys contact list and are cached locally (never numbers) so Claude can mention people by name instead of by raw number. Any number that still has to be shown anywhere (an ambiguous name match, a group roster with an unknown member) is masked to the last 4 digits, built masked from the start.
- `"all"` in `reply`'s `mentions` array expands to every current group participant, fetched live, so it works even for members with no saved contact name.
- Roster access (seeing who is in a group, needed for `all`) is a separate grant from replying in the group at all, off by default.
- A new terminal-only checkbox wizard (`bun scripts/access.ts wizard`, `@inquirer/prompts`) shows the account's most recently active groups and DM contacts for bulk setup. It never runs through any AI model, on purpose, so that guarantee is verifiable rather than asserted.
- Explicitly removing a contact now also forgets their cached name locally, not just their access.

Full detail in the updated ACCESS.md / USAGE.md.

## One thing worth your input

Contact names are cached in plaintext at `~/.whatsapp-channel/contacts.json` (never numbers, see above) so mentions and group rosters can resolve by name. Happy to make this opt-in if that's a better default for the plugin's threat model, rather than assuming it is.

## Test plan

- [x] `bun test` (140 pass, 13 skip pre-existing/Windows-only, 0 fail)
- [x] Typecheck and bundle clean
- [ ] Manual smoke test of the checkbox wizard and `@all` in a live group (automated coverage stops at the interactive prompt boundary - noted in the code)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "whatsapp-claude-channel",
       "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.15.3",
+      "version": "0.16.0",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "whatsapp-claude-channel",
   "displayName": "WhatsApp Channel for Claude Code",
   "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-claude-channel:access.",
-  "version": "0.15.3",
+  "version": "0.16.0",
   "author": {
     "name": "Rich627"
   },
--- a/ACCESS.md
+++ b/ACCESS.md
@@ -53,13 +53,14 @@ Group JIDs end in `@g.us`. To find one, add the linked device to the group — t
 
 With the default `requireMention: false`, the server responds to every message. Pass `--mention` to require @mention, or `--allow jid1,jid2` to restrict which members can trigger it. Pass `--roster` to also grant roster access (see below) — off by default, same as everything else.
 
-Running `group add` again on an already-configured group **merges**, it doesn't start over: any flag you don't pass this time keeps whatever was already set. Adding `--roster` to a group that already has `--mention` on doesn't reset `--mention` back off — only the flags you actually pass change anything.
+Running `group add` again on an already-configured group **merges**, it doesn't start over: any flag you don't pass this time keeps whatever was already set. Adding `--roster` to a group that already has `--mention` on doesn't reset `--mention` back off — only the flags you actually pass change anything. To explicitly turn `--mention` or `--roster` back off (rather than just never setting them), pass `--no-mention` / `--no-roster` — passing both a flag and its negation at once is refused, not silently resolved one way.
 
 ```
 /whatsapp-claude-channel:access group add 120363424405607157@g.us
 /whatsapp-claude-channel:access group add 120363424405607157@g.us --mention
 /whatsapp-claude-channel:access group add 120363424405607157@g.us --allow 886912345678@s.whatsapp.net
 /whatsapp-claude-channel:access group add 120363424405607157@g.us --roster
+/whatsapp-claude-channel:access group add 120363424405607157@g.us --no-roster
 /whatsapp-claude-channel:access group rm 120363424405607157@g.us
 ```
 
@@ -103,6 +104,7 @@ Both fail with a clear error if the group's `roster` flag isn't granted.
 - **Contacts**: pick which ones can message Claude — added straight to the allowlist, no second question.
 - Archived groups are skipped by default (pass `--include-archived` to include them). Anything already configured, or outside the top 5/10, isn't shown here — add it individually later with `group add`/`allow`, or just ask Claude (it already knows the name from context).
 - Ctrl-C cancels cleanly at any point; nothing is written until every question on screen has been answered.
+- A group the wizard adds gets `requireMention: true` (only reply when addressed) — a more cautious default than `group add`'s own CLI default of `false` (reply to everything), deliberately: the wizard is the guided path for a less technical setup, the CLI is for someone already comfortable with explicit flags. Change it after the fact with `group add --mention` or `--no-mention`.
 
 Group/contact names and recency come from caches (`~/.whatsapp-channel/groups-meta.json`, `dm-activity.json`) that only the running server writes — automatically, as WhatsApp reports chat activity, no manual step needed once the account has connected at least once. If it's never connected yet, the wizard has nothing to show; pair it first.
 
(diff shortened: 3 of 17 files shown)


Rich627 (changes requested):
Thanks — this is a really well-thought-out PR. I ran the suite locally from your branch (153 pass, 0 fail) and read through all of it. The trust-boundary work in particular is excellent: `resolveByName` matching only the owner-saved `.name` and never self-reported `.notify` (with the display-name-hijack rationale written down), `group_roster` treating number-shaped notify values as no-name, and the `{input, jid}` mention pairs fixing the LID-resolution text-match drop. The terminal-only wizard making the "no AI involved" guarantee verifiable rather than asserted is a design I'm happy to have.

Answers to your questions first:

- **`@inquirer/prompts`: accepted.** The wizard is terminal-only and hand-rolling a checkbox UI isn't worth the maintenance. I'll note the dependency-policy exception in the repo docs.
- **Plaintext `contacts.json`: keep it as the default, no opt-in needed.** `~/.whatsapp-channel/` already holds the full message log and the Baileys auth keys at 0700/0600 — a name cache in the same directory isn't a meaningful escalation, and opt-in would gut mention-by-name for everyone.

Two real bugs I'd like fixed before merge:

1. **`refreshGroupsMeta` drops `lastActivityAt` when it rewrites an entry.** When a group's name or member count changes, the new object carries `name/memberCount/archived/updatedAt` but not `lastActivityAt: existing?.lastActivityAt` — so every `list_groups` call (or connect-time warm) wipes the recency data the wizard ranks by, until the next `chats.update` happens to restore it. One-line fix.

2. **`dmActivity` entries keyed under a `@lid` JID are never migrated.** `recordLidMapping` migrates `contactsMap` via `migrateContactKey`, but a DM whose activity was recorded before the lid→phone mapping was learned stays keyed by the lid forever, while later writes go under the phone key. Result: the same person can appear twice in the wizard's top-10 (once by name, once as a masked LID), and `remove <jid>` only deletes the phone-keyed activity entry — so "forget" doesn't fully clear the recency cache, which matters for exactly the privacy story this PR is telling.

Two smaller asks:

3. **There's no way to revoke `roster` (or `requireMention`) from the CLI.** With the merge semantics, `group add` computes `roster: roster || existing.roster`, so once granted, the only way to turn off the single most privacy-sensitive flag in this PR is `group rm` + re-add, losing `allowFrom` and the mention setting along the way. Please add explicit `--no-roster` / `--no-mention` flags (Node's `parseArgs` doesn't auto-negate, so they'd be their own options).

4. **The wizard hardcodes `requireMention: true` for groups it adds, while `group add` defaults to `false`.** The safer default is arguably right for bulk setup, but it's inconsistent between the two paths and ACCESS.md doesn't mention it — please document it in the wizard section (or align the two, your call which way).

Housekeeping:

- **CI red is purely formatting**: 9 files need this repo's `trunk fmt`, plus three markdownlint MD060 (table column style) hits in USAGE.md.
- **Version bump**: please bump both `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json`'s `plugins[0].version` to `0.16.0` (leave marketplace's top-level `version` alone) — repo rule, new features = minor.

With those addressed I'm happy to merge.

Rich627 (changes requested):
Thanks for tackling this — and for addressing two items from the #16 review along the way (the device-suffix match key now derives from the normalized JID, and the ranking normalizer mirrors the real Baileys behavior). The dependency is fine: `@inquirer/prompts` is approved. The direction is good, but the review turned up problems in each of the PR's three core guarantees (masking, forget-on-remove, the AI-free wizard), plus some correctness issues in the name-mention path. Requesting changes on the following.

**Security / privacy (blocking):**

1. **The wizard can silently revert live `access.json` writes** (`scripts/access.ts` ~448). It snapshots `access.json` up front, blocks on interactive prompts for unbounded time, then writes the whole stale snapshot back. Any server write in between — a pairing approval appended to `allowFrom`, pending-code creation/pruning — is reverted: the just-approved sender is dropped (their pairing code already consumed) with no error anywhere. The one-shot commands' "few-ms window" reasoning doesn't hold for an interactive command. Please re-`load()` at save time and apply only the fields the wizard changed.

2. **Wizard labels trust attacker-controlled names** (`scripts/ranking.ts` ~94). DM candidates are labeled `name || notify`, and `.notify` is the sender's self-chosen push name — which `contacts.ts` itself documents as untrusted. Since `dm-activity.json`/`contacts.json` are populated by ungated Baileys events even for senders the access gate rejected, an unknown number that DMs once with push name "Mum" shows up top-of-list as "Mum", and one tick grants it full DM access — under a banner asserting the decision was AI-free. Please mark non-`.name` labels as self-reported and/or show the masked number alongside.

3. **Forget-on-remove doesn't hold while the server is running** (`scripts/access.ts` ~529). `remove` deletes from the files, but the running server holds both maps in memory (loaded once, never re-read) and every save serializes the whole map — so the next `contacts.upsert` or LID migration writes the forgotten entry straight back. Server-up is the normal state, so the guarantee is normally void.

4. **Unconditional plaintext persistence, including in STATIC mode, with no purge path for strangers** (`server.ts` ~2845). The `contacts.*`/`chats.*` handlers persist the entire synced address book and a recency log of every rejected sender; `WHATSAPP_ACCESS_MODE=static` guards `saveAccess` but not these; and `access.ts remove` `die()`s for non-allowlisted JIDs, so someone who was never allowlisted can never be purged. On your open question: yes, make the contacts cache opt-in — but the STATIC bypass and the missing purge path need fixing regardless.

**Correctness (blocking):**

5. **Name mentions are a silent no-op in LID-addressed groups** (`scripts/mentions.ts` ~58). The name branch ships the contacts-map key (phone-form JID) verbatim, skipping the prefer-LID logic the bare-number branch applies — the @ renders as inert text, nobody is notified, the tool reports success. The numeric path resolves the same person correctly, so the new recommended path is strictly worse than the one it replaces.

6. **An unresolvable name is shipped as `<name>@s.whatsapp.net`** (`scripts/mentions.ts` ~93). `group_roster` displays notify-derived names, but `resolveByName` ignores `.notify`, so passing such a name (or any typo) falls through the numeric branch into a nonsense JID and a silent failed ping. Please raise, like the ambiguous-name case already does.

7. **`group_roster` masks LID digits, not phones** (`server.ts` ~2089). Identity is resolved only via the passively-populated lidMap, ignoring the `phoneNumber` field Baileys returns on the same participant — so every member who's never spoken renders as a masked LID, and saved contacts aren't recognized. Please key the lookup off `p.phoneNumber` (and consider seeding the lid-map from roster fetches).

8. **Masking is two renderers, not a property of the identity layer** (`server.ts` ~2670). Inbound meta, `unreplied`, and `catch_up` still ship raw numbers on every message, while the new reply-schema text asserts a raw number never needs to appear. Suggest one `displayIdentity(jid)` resolver routed through persist/notify/unreplied/catch_up, keeping raw JIDs only in addressing fields.

**Also, please:**

- Consolidate the two divergent roster gates (`reply`'s `@all` at ~1713 vs `group_roster`'s hand-rolled check at ~2079 — the latter lacks the `isAllowedJid` branch) into one `assertGroupCapability(chat_id, 'roster')`.
- `looksLikeNumber` misses `.` and `/` separators (verified by execution: `"61.403.911.675"` → false), so such push names print raw in the roster and wizard.
- A JSON-number/null entry in `mentions` now throws in `isReservedAllToken` (`entry.trim()` on uncoerced input) and fails the whole reply — the `String(entry ?? "")` coercion that guards `normalizeMentionJids` was dropped on this path.
- `refreshGroupsMeta` never prunes left groups, so the wizard keeps offering them forever.
- Perf: `chats.update` does a synchronous whole-file rewrite of dm-activity/groups-meta per inbound message over unbounded maps — please batch/debounce; and `reply()` reads `access.json` up to three times per call.
- Dedup: four identical load-or-`{}` helpers and three copies of the atomic-write sequence in `access.ts`; byte-identical `chats.upsert`/`chats.update` bodies; `GroupMeta` redeclared instead of imported.

**Process suggestion:** since this branch builds on #16 and reworks the same mention logic, let's land #16 first (it has its own pending review items) and rebase this on top — reviewing both in flight against each other will just churn.

MrTig-afk:
Pushed 7 of the 8 items - the ones independent of the scripts/mentions.ts -> lib/mentions.ts move happening in #16:

1. **Wizard reverting live access.json writes** - re-loads at save time and applies only the wizard's own changes, instead of writing back the pre-prompt snapshot.
2. **Untrusted `.notify` labels** - a name-only-via-`.notify` DM candidate is now labeled `"Name (unverified) - •••••1234"` instead of showing identically to a saved contact.
3. **Forget-on-remove while the server is running** - contactsMap/dmActivity now reload from disk immediately before every mutate-then-save path (contacts.upsert/update, chats.upsert/update, LID migration), so an external `remove`/`forget` survives instead of being overwritten by the next sync event.
4. **Plaintext persistence + no purge path** - `contacts.json`/`dm-activity.json` writes are now gated behind `WHATSAPP_CACHE_CONTACTS=1` (off by default), also off unconditionally in STATIC mode. Added `bun scripts/access.ts forget <jid>`, which purges a cached name/activity entry with no allowlist requirement (unlike `remove`) - for a stranger who was never granted anything to remove.
5. **LID no-op for name mentions** - a name-resolved mention now prefers the LID form when known, same as the numeric path already did.
6. **Unresolvable name -> nonsense JID** - now raises an error instead of silently shipping `"<name>"@s.whatsapp.net`.
7. **`group_roster` masking LID digits** - now keys off `groupMetadata()`'s own `phoneNumber` field directly instead of only our passively-populated lidMap, and feeds it back into lidMap for other lookups.

**Item 8 (the mentions.ts device-suffix fix) is deliberately not here** - per your process suggestion, I'll pick it up once #16 lands and rebase onto it, rather than duplicating the fix on the pre-move file path.

All tests/typecheck/prettier clean (155 pass, 0 fail).

Rich627 (changes requested):
Re-reviewed the rebased branch. I verified every blocking item from both earlier reviews against the actual code (grep + execution, not just the commit messages) — **all 12 are genuinely fixed**, and nothing regressed in the rebase: `lastActivityAt` preserved on rewrite, dm-activity LID migration, `--no-roster`/`--no-mention` (with the both-flags guard), wizard default documented, reload-at-save, `(unverified)` labels, reload-before-mutate forget durability, `WHATSAPP_CACHE_CONTACTS` opt-in + STATIC guard + `forget`, name mentions preferring LID, unresolvable names raising, roster keyed off `phoneNumber`, and the device-suffix fix carried through from #16. Version 0.16.0 is right in both files, trunk and the version-pairing check are green, and the full suite passes locally on macOS (168/168). Nice work.

**One thing still blocks merge: the CI Test job is red.** All 6 failing tests are the wizard ones, and the failure mode is uniform — the spawned wizard child hangs and bun's 5s default test timeout kills it. Root cause: `@inquirer/prompts`' checkbox doesn't receive keystrokes over piped, non-TTY stdin on Linux (your own comment in `access.test.ts` cites the Inquirer issue — it just bites harder there than on macOS, where these pass). Any of these is fine by me: drive the prompts through a PTY, give the wizard a non-interactive input path for tests, or skip the prompt-driving tests on CI while keeping the ranking/filtering coverage that already lives in `ranking.test.ts`.

**One small fix to fold in while you're there:** a non-string entry in `mentions` (a JSON number, or null) still fails the whole reply — `isReservedAllToken` calls `entry.trim()` directly, and at the `reply` call site `isAllToken` runs on the raw array *before* `normalizeMentionJids`'s `String(entry ?? "")` coercion (server.ts ~1737). Verified by execution: `isReservedAllToken(42, {})` throws `entry.trim is not a function`. Coercing at the top of `isReservedAllToken` mirrors the guard the other path already has.

The remaining items from the earlier "Also, please" list (`looksLikeNumber` missing `.`/`/` separators, the duplicated roster gates, left groups never pruned from the wizard candidates, the byte-identical `chats.upsert`/`chats.update` bodies) — I'll spin those out as follow-up issues rather than hold this PR on them. The write-path change-detection you added already covers the perf concern well enough.

Heads-up on coordination: **0.16.0 collides with what #18 was asked to bump to** — whichever PR merges second needs another bump. This one looks closer to landing, so #18 will likely take 0.17.0.

Rich627 (approved):
All conditions met — CI green, both review rounds fully addressed. Merging. Follow-ups tracked in #19.

PR #18, opened 2026-08-22
Multi-terminal WhatsApp sync

MrTig-afk:
## What & why

Right now, opening a second Claude Code terminal with this plugin loaded just refuses to work while another terminal already holds the WhatsApp connection - WhatsApp only allows one real linked-device connection at a time. This PR adds:

- A singleton lock + IPC socket relay: the first process to start becomes the primary and holds the real connection; every later process becomes a secondary that relays its tool calls (`reply`, `react`, etc.) and re-emits the primary's inbound-message notifications, over a local, same-machine, token-authenticated socket. From inside Claude Code, both terminals behave the same.
- Auto-promotion: if the primary exits (clean or crashed), a connected secondary detects it, retries, and promotes itself to primary automatically - no restart needed.
- Same-machine only, two terminals, by design (not cross-machine, not 3+).

## How I tested it

Full local suite (`bun test`, `prettier --check`, `tsc --noEmit`, `bun build`) - all clean, no failures.

Also live-tested with two real terminals against a real (throwaway, pairing-code-linked) WhatsApp account, not just the unit tests:
- **Promotion**: killed the primary's OS process directly mid-session. The secondary's `.server.lock` took over automatically within seconds - no restart, no manual intervention.
- **Relay**: issued a `reply` call from the SECONDARY terminal specifically. It was relayed through the primary's connection and the message was confirmed received on the linked phone.

## Checklist

- [ ] I ran `trunk fmt` and `trunk check` and it passes. (Not run locally - `trunk` isn't installed in this environment. `bunx prettier --check`, `tsc --noEmit`, and `bun test` were all run and are clean; deferring `trunk check` itself to CI.)
- [x] **Version bumped in BOTH files to the same new version** (0.15.2 -> 0.15.3, feature):
  - [x] `.claude-plugin/plugin.json` -> `version`
  - [x] `.claude-plugin/marketplace.json` -> `plugins[0].version` (the inner one)
- [x] No runtime state / secrets committed (nothing from `~/.whatsapp-channel/`).
- [x] Touched `server.ts` connection lifecycle and the singleton lock - invariant preserved: exactly one process ever holds the real WhatsApp connection at a time; a secondary never attempts its own connection while a primary is reachable, and promotion only happens after the singleton lock is confirmed free.


Rich627 (changes requested):
Thanks — this is solid work. The core design is right (secondaries never touch Baileys; the only path to `connectWhatsApp()` is winning the singleton lock), the token handling is careful (fresh per-start, 0600, written only after a successful bind, fail-closed), and the tests are genuinely good — real `server.ts` children over real MCP stdio against a wire-accurate fake primary, with regression rationale attached. Happy to merge this once a few things are addressed:

**1. Rebase + version re-bump.** `main` merged #16 after you branched, so `main` is already at 0.15.3 and `server.ts` now conflicts (the mention-extraction code moved to `lib/mentions.ts`). Please rebase onto current `main` and re-bump to **0.16.0** (feature → minor). I'd like to take one more look at the resolved `server.ts` after the rebase, since the conflict sits right on top of #16's changes.

**2. Comments referencing documents that aren't in the repo.** There are 30+ comment references to `PRD §…`, `FR1–FR6`, `T03/T04/T05`, `.pipeline/spec.md`, `.pipeline/review.md`, `.pipeline/test-results.md` — none of these exist here, so future readers can't resolve any of them. Either commit the spec into the repo or reword those comments to be self-contained. Same for the five `ponytail:` marker comments — please drop the tag (the content of those notes is worth keeping).

**3. Document the double-delivery behavior.** Every inbound message is delivered to the primary's session *and* broadcast to each secondary's session, and both sessions are Claude instances that may act on it — so one WhatsApp message can get two independent replies. I'm fine shipping with this behavior, but USAGE.md's "Session conflicts" section must state it explicitly so users aren't surprised.

Minor, non-blocking:

- On Windows, the 3s lock-retry tick synchronously spawns a PowerShell CIM query forever while degraded — your comment acknowledges it; fine for now.
- `.role-<pid>` files persist after a crash (only cleaned on graceful shutdown); consider pruning dead-pid role files at startup.
- Startup race: the initial `ipcRelay = await connectToPrimary(onRelayLost)` can overwrite the `null` set by an `onRelayLost` that fired before the await resumed, leaving a dead relay marked live for one retry tick. Self-heals in ~2s; a note or a guard is enough.

Rich627:
Re-reviewed after your update — all three asks are done and verified: the internal-reference comments are fully reworded (I checked the reword commit is comment-only, no logic drift), the double-delivery behavior is documented clearly in USAGE.md, and the rebase kept #16's mention integration byte-identical. CI is green and the suite passes locally here too (100/100). Thanks for the fast turnaround.

Two mechanical things before this can land, both caused by #17 merging just now:

1. **Rebase onto current `main` once more.** I test-merged locally: `server.ts` conflicts in three spots, all in the same area — the import block (your IPC imports vs #17's expanded `lib/mentions` imports) and two regions of the `tools/list` array, where your degraded-ternary re-indent of the whole tool list meets #17's rewritten `reply.mentions` description and its new `group_roster` tool. When resolving, take #17's tool descriptions/entries as the content and re-apply your ternary wrap around them — and note `group_roster` is a new tool that needs to be inside your non-degraded branch.

2. **Bump to `0.17.0`** (both files, as before) — `main` is now `0.16.0` via #17.

No re-review needed beyond a quick look at the resolved `tools/list` — everything else is already verified.

Rich627:
Heads-up: #20 landed while you were rebasing, which re-conflicted only the two version files (main went to 0.16.1) and that in turn kept CI from running. Since it was trivial, I resolved it via maintainer push — rebased your branch onto current main keeping your 0.17.0, zero changes to your commits' content.

Verified before merging: your #17 conflict resolution is exact (the tools/list content is character-identical to main's inside your degraded ternary, `group_roster` correctly in the non-degraded branch), the reword commit is comment-only, the double-delivery section in USAGE.md covers what I asked, and the full suite passes at the final tip (214/214) with CI green. Nice work on this one — the test rigs especially.

Rich627 (approved):
All conditions verified across both review rounds — merging at 0.17.0.

PR #22, opened 2026-08-23
feat(notifications): default-on proactive notifications + update notice

MrTig-afk:
## Summary

- Claude now proactively tells you about a pairing code, a "paired and connected" confirmation, a role change (primary/secondary/reconnecting), and an inbound message — instead of waiting for its next natural reply. Default on, opt out per terminal with `WHATSAPP_QUIET=1`.
- One-time update notice on a returning terminal after a version bump, tracked per-account in `.last-seen-version`. Closes the gap where the checkbox wizard (#17) shipped with no way for an already-paired user to learn it exists — the 0.18.0 changelog entry surfaces it on next restart.
- Initial role a terminal boots into is never announced as a "change" — only real transitions are.

## Test plan

- [x] `bun test` — 205 pass / 0 fail / 13 skip (218 tests, 12 files)
- [x] `bunx tsc --noEmit --skipLibCheck --esModuleInterop --target es2022 --module esnext --moduleResolution bundler --types node,bun server.ts scripts/ipc.ts scripts/ipc.test.ts scripts/ipc-relay.test.ts scripts/ipc-promotion.test.ts` — clean
- [x] `bunx prettier --check server.ts scripts/ipc-promotion.test.ts ACCESS.md` — clean
- [x] New coverage in `scripts/ipc-promotion.test.ts`: initial-role-not-announced regression, role-change notifications on disconnect/reconnect/promotion, `WHATSAPP_QUIET=1` suppression, update-notice fires with older recorded version (asserts the wizard instructions specifically appear in the notice text), update-notice stays quiet when version matches


Rich627 (changes requested):
Thanks — and the diagnosis behind this is right. An already-paired user has no path to
learn anything: their interface is WhatsApp, not the plugin, `plugin update` is silent,
and README changes only reach people who go back to GitHub. A feature that needs a
terminal command is invisible to them forever. That's a real hole and I'm glad you
spotted it.

I want to split the PR, though, because the two halves have very different
justifications. Details below, but the short version: **the notification half I want as
it stands; the update-notice half I'd like moved out of `server.ts` and into the
SessionStart hook that already exists for exactly this.**

## The notification half — yes, merge this

This one earns its place in `server.ts` because a role change is a runtime event that
only that process knows about. Without it, a terminal that quietly became a secondary or
lost the primary connection looks identical to one that's working — and this repo has
been bitten by "looks alive, isn't" more than once (a half-dead socket where groups go
silent while DMs still flow; expired auth where Baileys is happy, messages arrive, and
every reply is "Login expired"). #18 introduced a new flavour of that silent state, so
announcing transitions is closing a gap #18 opened.

Things I checked and am satisfied with:

- **Ordering is safe.** `announceUpdateIfNeeded()` runs at module tail (server.ts:3979),
  well after `await mcp.connect(new StdioServerTransport())` (server.ts:2950) — the
  notice can't be emitted into a transport that isn't up yet.
- **`pastInitialRole` covers all three startup branches.** `becomePrimary()`
  (server.ts:996), the secondary boot path (server.ts:3971), and the degraded
  `startRetryLoop()` (server.ts:3977) each make the *first* `writeRoleFile()` call for
  the process, so none announce. Putting the flag inside `writeRoleFile()` instead of at
  each call site is the right call — the earlier fix's failure mode (only primary
  special-cased, so every ordinary secondary boot announced) is exactly what per-call-site
  opt-outs invite.
- **`localeCompare(..., {numeric: true})` is correct**, and choosing `0.9.0` over `0.1.0`
  for the fixture is the difference between a test that catches the string-compare bug and
  one that doesn't.

## The update-notice half — right problem, but there's already a place for this

`hooks/hooks.json` wires a SessionStart hook to `hooks-handlers/session-start.sh`. It
already runs at every session start, already inspects `~/.whatsapp-channel/` state, and
already injects text into the session via `additionalContext`. Its final branch — the
fully-configured case — currently says nothing but "WhatsApp channel is fully configured
and ready."

That `else` branch is where an update notice belongs. Cost there: read one more file, one
more `if`, a few lines of bash.

Cost of the current approach:

- a hand-maintained `CHANGELOG` array (server.ts:451) living in a ~3900-line file that
  `CLAUDE.md` explicitly flags as having danger zones
- another state file, `LAST_SEEN_VERSION_FILE` (server.ts:98)
- a maintenance obligation on every future release ("does this version need an entry?"),
  where forgetting fails silently

There's also an inconsistency worth naming: the notice ships over
`notifications/claude/channel`, which — as you established yourself in #15 — is a Claude
Code extension, not standard MCP. So under the non-Claude-Code clients you spent #14
supporting, the update notice silently never arrives.

In fairness: hooks are Claude Code-only too, so that path doesn't reach those clients
either. Which is the point — routing through `server.ts` buys no extra coverage here, it
just costs more. The split I'd draw is by the shape of the event: a role change is a
runtime event and must come from the server; a version bump is a session-start-shaped
check against a file, with nothing to do with the WhatsApp connection at all.

Happy to take this as two PRs, or as one with the notice reworked — whichever is less
work for you.

## Carries over either way

**`bun scripts/access.ts wizard` is repo-relative, and its audience is precisely the
people who aren't in the repo.** This one applies to the "paired and connected" message
(server.ts:3836) regardless of what happens to the update notice, and to the notice
wherever it ends up. For a marketplace install the plugin lives under
`~/.claude/plugins/cache/whatsapp-claude-plugin/...`, so pasting that line gets
`error: Module not found "scripts/access.ts"`.

The repo already has the idiom — `skills/doctor/SKILL.md:19` and `skills/setup/SKILL.md:91`
use `bun "${CLAUDE_PLUGIN_ROOT}/scripts/..."` — and this PR already proves `server.ts` can
resolve its own location, in the `PLUGIN_VERSION` IIFE. So it can interpolate a real
absolute path with no guessing:

```ts
const WIZARD_CMD = `bun ${JSON.stringify(join(import.meta.dir, "scripts", "access.ts"))} wizard`;
```

(The relative form in README.md / ACCESS.md is fine and should stay — those are read while
sitting in the repo. It's only the runtime-delivered strings that have the wrong audience.)

## Minor, no action needed

- **Only one terminal ever sees the update notice.** `.last-seen-version` is per-account
  and written unconditionally on boot, so after an update the first terminal up consumes
  it and the rest stay quiet — and that isn't necessarily the one the user is looking at.
  Note this is *not* fixed by moving to the hook; the same race exists per session. Once
  per account is a defensible reading, I just want it to be a decision rather than a
  consequence of where the write landed, and written down wherever the notice ends up.
- **`notifyRoleChange()` doesn't compare against the previously written role**, so a
  flapping relay produces an alternating "Lost the primary…" / "…now a secondary" stream.
  Each is a genuine transition so it isn't wrong; if it turns out noisy in practice, a
  `lastRole` compare inside `writeRoleFile()` is the one-line fix.

Rich627:
Thanks for the quick shellcheck fix — almost there. The latest run still fails Trunk, but it's down to one shfmt formatting complaint in `hooks-handlers/session-start.sh`:

```diff
-	if [[ -n "${notice_json}" ]]; then
+	if [[ -n ${notice_json} ]]; then
```

shfmt runs in simplify mode here, so it strips the (unnecessary) quotes inside `[[ ]]`. Running `trunk fmt hooks-handlers/session-start.sh` produces exactly this change. Everything else looks good — once CI is green this is ready to merge from my side.

Issue #2, opened 2026-05-11
Posible bug


justimartinez:
## Symptom

WhatsApp inbound messages arrive at the bridge (Baileys side) and the typing indicator is sent, but Claude never processes the message and never replies. Plugin process stays alive — silent failure, no visible error.

## Root cause (confirmed via --debug-file)

Claude Code 2.1.x (research preview) enforces a server-side allowlist of approved channel plugins. This plugin is third-party (not in `claude-plugins-official`), so its `mcp.notification('notifications/claude/channel', ...)` calls are silently dropped.

Exact line from Claude Code debug log:

[DEBUG] MCP server "plugin:whatsapp:whatsapp": Channel notifications skipped: plugin whatsapp@whatsapp-claude-plugin is not on the approved channels allowlist (use --dangerously-load-development-channels for local dev)

## Versions

- Claude Code: 2.1.133 / 2.1.138 (both reproduce). Worked fine on 2.1.119.
- whatsapp-claude-plugin: 0.6.2
- OS: Windows 11

## What we tried (all failed)

1. Cache wipe + clean restart — same error.
2. Rollback Claude Code to 2.1.133 — allowlist already enforced.
3. `allowedChannelPlugins` setting in user `~/.claude/settings.json` — ignored. Setting only takes effect in managed settings (Team/Enterprise plans).
4. `--dangerously-load-development-channels` flag — does not lift the allowlist for `--channels plugin:...` format.

## Reference

- Channels docs: https://code.claude.com/docs/en/channels
- `allowedChannelPlugins` introduced in Claude Code v2.1.84 (March 26, 2026). Strict enforcement appears to have rolled out between 2.1.119 (April 24) and 2.1.133 (May 7).

## Possible workarounds

- Bypass `notifications/claude/channel` and write directly to `~/.claude/channels/<name>/inbox/` (file-based polling).
- Petition Anthropic to add this plugin to the official allowlist.

Rich627:
Thanks for the detailed write-up. The allowlist mechanism you described is real (confirmed in the Claude Code 2.1.x binary — `isChannelAllowlisted` / `getChannelAllowlist` exist, and the error string you quoted matches verbatim). The plugin has been approved by Anthropic; the sync to the runtime allowlist (`tengu_harbor_ledger`) is the gating factor on Anthropic's side, not something this repo can patch directly.

A couple of things would help confirm we're looking at the same bug before I escalate upstream:

1. The README launch command is:

   ```
[...]

PR #24, opened 2026-08-24
add in-session checkbox access review to the access skill

MrTig-afk:
## Summary
- Adds `review` and `manage` subcommands to `skills/access/SKILL.md`, an in-session checkbox equivalent of the terminal `wizard` command (Claude Code only) using `AskUserQuestion` as the human-in-the-loop gate.
- `review` grants access to new groups/contacts; `manage` revokes existing access (empty selection = remove nothing).
- Backed by two new pure functions in `scripts/ranking.ts`: `listConfiguredGroups`, `listConfiguredDms`.
- Terminal `wizard`'s "zero AI involvement" guarantee is untouched — `scripts/access.ts` has zero diff.

## Test plan
- [x] `bun test scripts/ranking.test.ts` — 37 pass, 0 fail
- [x] `bun test` (full suite) — 228 pass, 13 skip, 0 fail
- [x] `bunx tsc --noEmit` on the changed files — clean
- [x] `bunx prettier --check` on the changed files — clean
- [x] Manual read-through confirming dispatch prose in SKILL.md follows the existing subcommand pattern (numbered steps, exact paths, exact JSON shape)

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "whatsapp-claude-channel",
       "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.18.1",
+      "version": "0.19.0",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "whatsapp-claude-channel",
   "displayName": "WhatsApp Channel for Claude Code",
   "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-claude-channel:access.",
-  "version": "0.18.1",
+  "version": "0.19.0",
   "author": {
     "name": "Rich627"
   },
--- a/.gitignore
+++ b/.gitignore
@@ -2,3 +2,4 @@ node_modules/
 bun.lock
 package-lock.json
 CLAUDE.md
+.claude/
--- a/scripts/access.test.ts
+++ b/scripts/access.test.ts
@@ -181,6 +181,54 @@ describe("allowlist", () => {
     expect(res.code).toBe(0);
     expect(res.out).not.toContain("Forgot");
   });
+
+  // One contact can be allowlisted twice, under its @lid form AND its phone
+  // form, and both resolve to the same contacts.json / dm-activity.json key.
+  // Revoking one form must not wipe the cache the surviving grant still
+  // relies on (PR #24 review, #3).
+  test("revoking one form of a doubly-allowlisted contact keeps the shared cache", () => {
+    const dir = freshStateDir();
+    writeLidMap(dir, { "184710990000999@lid": "61403911675@s.whatsapp.net" });
+    writeContacts(dir, { "61403911675@s.whatsapp.net": { name: "Akash" } });
+    writeDmActivity(dir, { "61403911675@s.whatsapp.net": 1000 });
+    run(dir, "allow", "184710990000999@lid");
+    run(dir, "allow", "61403911675@s.whatsapp.net");
+
+    const { out } = run(dir, "remove", "184710990000999@lid");
+    expect(out).toContain("Kept their cached name");
+    expect(out).not.toContain("Forgot their cached name");
+    // The surviving grant still works, so the name behind it must survive too.
+    expect(readContacts(dir)["61403911675@s.whatsapp.net"]).toEqual({
+      name: "Akash",
+    });
+    expect(access(dir).allowFrom).toEqual(["61403911675@s.whatsapp.net"]);
+  });
+
+  test("nothing cached: the surviving grant is kept quietly, no invented name claim", () => {
+    // Allowlisted by JID, never DMed - so there is no cached name to keep,
+    // and saying one was kept is as wrong as saying one was forgotten.
+    const dir = freshStateDir();
+    writeLidMap(dir, { "184710990000999@lid": "61403911675@s.whatsapp.net" });
+    run(dir, "allow", "184710990000999@lid");
+    run(dir, "allow", "61403911675@s.whatsapp.net");
+
+    const { out } = run(dir, "remove", "184710990000999@lid");
+    expect(out).toContain("Removed 184710990000999@lid.");
+    expect(out).not.toContain("cached name");
+  });
+
+  test("revoking the LAST form of that contact does forget them", () => {
+    const dir = freshStateDir();
+    writeLidMap(dir, { "184710990000999@lid": "61403911675@s.whatsapp.net" });
+    writeContacts(dir, { "61403911675@s.whatsapp.net": { name: "Akash" } });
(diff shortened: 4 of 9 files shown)


Rich627:
@codex review

Rich627 (commented):
Thanks for this — the direction is great (in-session checkbox review beats bouncing users out to the terminal wizard), and the `groupLabel`/`dmLabel` extraction into shared helpers verified as behavior-preserving. A detailed pass turned up a few things to fix before merge, roughly in order of severity:

## Blocking

**1. Missing version bump.** `.claude-plugin/plugin.json` and `marketplace.json` (`plugins[0].version`) are both still `0.18.1` on this branch. Repo rule: every user-facing change bumps both, or `plugin update` silently no-ops for installed users. This adds new subcommands, so it should be a minor bump → `0.19.0` in both files.

## Bugs

**2. `manage` pagination is a dead end (SKILL.md ~L345).** The prose says "cap each list at 4 … re-running `manage` covers the next batch", but `listConfiguredGroups`/`listConfiguredDms` sort alphabetically with no pagination state, and `manage` only ever *removes* entries. A user with 6 `allowFrom` entries sees the same first 4 forever — entries 5+ can never be revoked through `manage` without first revoking access they want to keep. (`review` is self-healing because approving removes candidates from the pool; `manage` isn't.)

**3. Cache wipe can hit a still-allowed contact (SKILL.md ~L379).** Step 8 removes only the exact selected `allowFrom` string, but step 9 deletes the LID-*resolved* key from `contacts.json`/`dm-activity.json` without checking whether another surviving `allowFrom` entry resolves to the same key. If someone is allowlisted under both `…@lid` and `…@s.whatsapp.net` and the user revokes just one form, the other grant survives but its cached name/recency is wiped — the contact degrades to a bare masked number everywhere, and step 10 reports them "forgotten" while they can still message Claude. Step 9 needs a "no other surviving entry resolves to this key" guard.

**4. Duplicate labels make selections ambiguous (`ranking.ts` ~L157).** `listConfiguredDms` deliberately returns both the `@lid` and phone forms of the same contact, but both get byte-identical labels (same resolved key → same `dmLabel` output), and AskUserQuestion selections come back *by label* — so the model can't map the tick back to a unique JID and may revoke the wrong form (or both). Ordering is also unstable (`localeCompare` of equal labels is 0). Two distinct groups sharing name + member count collide the same way. Labels need a disambiguator when they'd collide (e.g. last-4 / `[lid]`-`[phone]` tag).

## Security / privacy

**5. The label-overflow fallback drops the JID anchor (SKILL.md ~L289).** `review` step 5 says: if a label is too long, keep the name in the label and move the member count into `description` — which *replaces* the raw JID, the field the same paragraph designates as the one unspoofable identity anchor. Group names are attacker-controlled, so a long impersonating name forces exactly this branch and the human approves a grant with no JID shown anywhere. Truncate the label instead (or keep JID + count both in description).

**6. Raw JIDs in `description` vs the masking invariant (SKILL.md ~L286).** Putting every candidate's full unmasked number into the AskUserQuestion payload writes real phone numbers into the session transcript, which `mask.ts` explicitly tries to prevent — and the terminal wizard this mirrors never does it (JIDs ride as non-displayed checkbox values, `access.ts:409-424`). The anti-spoof rationale is real, but the model already knows each option's JID internally; `dmLabel`'s `(unverified) - •••••1234` marker may be enough for the human. Flagging as a deliberate-tradeoff question rather than a hard defect — but it currently also contradicts the same step's own label rule.

**7. Restated label rule drops the `looksLikeNumber()` guard (SKILL.md ~L270).** Steps restate the DM rule as "saved `.name` plain; otherwise `Notify (unverified) - •••••1234`; otherwise masked" — but the canonical `dmLabel` applies `looksLikeNumber()` to BOTH `.name` and `.notify` (`.notify` commonly defaults to the raw phone number). A model following the three-branch prose prints a full number into the label. Safer to just say "label exactly as `dmLabel` in `scripts/ranking.ts` computes" and drop the restatement.

## Smaller

**8. `[archived]` suffix missing from the prose spec (SKILL.md ~L264).** The spec says `Name  (N member(s))` but the canonical `groupLabel` appends `  [archived]` — and `manage` (where the tag matters most, to avoid revoking the wrong row) inherits the omission.

**9. Self-citations are off by one (SKILL.md ~L374 area).** The frontmatter insertion shifted the file, so the added line-range citations all point one line high — worst is `204-207` for "config/memory files are kept": the retention sentence is line 208, *outside* the cited range, so a model checking it could conclude no retention rule exists and delete `config.md`/`memory.md` on revoke. Also `access.ts:262` is a repo-relative pointer that resolves to nothing at a marketplace install. Citing by section/heading name (survives edits) would fix both.

**10. `listConfiguredGroups`/`listConfiguredDms` have no runtime callers.** Only the tests import them — the shipping behavior is the hand-followed prose, which already disagrees with the code (prose caps at 4; `access.ts` uses limits 5/10; `listConfigured*` take no limit). Suggestion that would dissolve #7/#8/#10 structurally: add read-only `candidates --json` / `configured --json` cases to `access.ts`'s existing switch and have the skill *execute* the canonical code instead of paraphrasing it.

Happy to discuss any of these — #2/#3/#4 are the ones I'd consider must-fix along with the version bump.

Rich627:
Verified the follow-up against all 10 review points — every one is addressed, mostly structurally rather than patched: the read-only `candidates`/`configured` commands plus `disambiguate()`/`clip()`/`groupAnchor()` dissolve the whole prose-drift class, the shared-key cache guard landed in `remove` itself, and the wizard's zero-AI guarantee is intact. Full suite passes locally (273/273). I pushed one tiny lint fix to your branch (MD040 — language tags on the two new command fences) to get CI green. Really nice work, thanks for turning this around so quickly and thoroughly 🙏

MrTig-afk:
Thanks for the lint fix and for going through it that carefully — I'd written the same MD040 patch locally and only spotted yours when the push bounced, so I dropped mine.

Heads up on what's next: I've got a branch with three fixes for problems I hit myself actually running the plugin day to day, and I'd like to open them as one PR once this merges.

1. **`WA:<role>` statusline segment never renders.** `statusline-role.ts` derives `PARENT_PID` from `process.ppid` and searches *downward*, but `statusLine` is a compound command, so it runs under a shell that is a **sibling** of the plugin wrapper, not an ancestor — the search can never reach it, `findServerPid` returns null, and it exits silently. Silent by design, which is why it has gone unnoticed since #18. Fix walks up the ppid chain (bounded, nearest-first) rather than scanning the whole process table, which would pick the wrong terminal's role with two terminals open.

2. **Update notice has no second chance.** It is emitted as `additionalContext`, so it lands in the model's context rather than the terminal, but the marker file advances unconditionally right after — if the model does not relay it that one session, the user never sees it. I hit this myself: the wizard announcement fired and I never saw it. Fix also compares the running version against what the marketplace clone advertises, no network call, and both notices now open with an explicit instruction to relay them.

3. **`wizard --revoke`.** There was no terminal path to revoke, only to grant. Reuses `listConfiguredGroups`/`listConfiguredDms` — the same functions the in-session `manage` reads through `configured` — and shares the cache-guard rule with `remove` via an extracted `revokeCachedIdentity()`.

Each is a real thing that bit me on my own machine, with the evidence in the issue write-ups on my fork. Happy to split them if you'd rather review them separately.

PR #26, opened 2026-08-25
Fix the statusline segment, make update notices reach the user, and give the wizard a revoke screen (0.21.0)

MrTig-afk:
Three issues I hit running the plugin day to day, plus three more turned up
on the way - two by code review, one by clearing the mess off my own machine. Rebased on top of the `whatsapp-channel` rename, so this
takes **0.21.0** — 0.20.0 is the rename itself.

## The three

**1. `WA:<role>` statusline segment never rendered** (since it landed in #18).
`statusline-role.ts` derived `PARENT_PID` from `process.ppid` and searched
*downward*. A `statusLine` setting is a compound command, so the script runs
under a shell the CLI spawned separately — a **sibling** of the plugin wrapper,
not an ancestor. The search could never reach it, and the miss is silent by
design, so it failed quietly for two days. It now climbs the ppid chain
(bounded, nearest-first) before searching down.

Two things surfaced while fixing it:

- The wrapper is matched on the plugin dir name, and the documented statusLine
  command *contains that name* (`bun <plugin-dir>/scripts/statusline-role.ts`).
  The shell was matching the wrapper pattern at the same BFS depth as the real
  wrapper, and a first-match search picked whichever row Win32_Process happened
  to list first. That turned a permanent failure into an intermittent one.
  Every candidate is now tried, and the one with a `server.ts` under it wins.
- Even fixed, a process-tree walk cannot tell this terminal's server from a
  sibling terminal's once a shared ancestor is in range — it would show
  someone else's role as yours. So `server.ts` now stamps the owning session
  (`CLAUDE_PID`) as line 2 of its role file, and the statusline takes the file
  whose owner is one of its own ancestors. Ownership is read, not inferred.
  Line 1 is unchanged, so older readers still work, and a server that has not
  restarted falls back to the tree walk.

  Side effect: a stale `.role-<pid>` from a crashed session can no longer be
  picked up, because a dead process is nobody's ancestor.

**2. The update notice had no second chance, and could only announce the past.**
It was emitted as `hookSpecificOutput.additionalContext` — the model's channel
— while `.last-seen-version` advanced unconditionally right after. If the model
did not relay it that one session, nobody ever saw it. And because it read this
plugin's own `plugin.json`, it could only ever say "you updated", never "an
update is waiting".

- Notices now ride on `systemMessage`, which goes to the user's terminal. No
  tokens, no dependence on the model repeating it, no "please relay this"
  preamble.
- A separate "newer version available" notice compares the running version
  against what the marketplace clone advertises (both already on disk, no
  network call). Deliberately **not** marker-gated, so it repeats until the
  user actually updates and goes quiet by itself when they do.
- The hook passes its own model-facing message through, so a session that gets
  a notice still briefs the model identically to one that does not.

**3. `wizard --revoke`.** The wizard could grant access but never take it back;
the only paths were `remove <jid>` one at a time or hand-editing `access.json`.
The revoke screen reuses `listConfiguredGroups`/`listConfiguredDms` — the same
functions the in-session `manage` reads through `configured` — so the terminal
and the chat cannot disagree about what is configured or how a row is labelled.
Empty selection removes nothing, and the cache rule is extracted to
`revokeCachedIdentity()` and shared with `remove` so the two cannot drift.

## The three found along the way

**A first-run notice showed the oldest release notes.** `CHANGELOG.slice(-1)`
takes the tail, but the array is newest-first — so a fresh install got a header
saying the current version over the *first* release's bullets. This is what bit
me: I was told about the wizard on an install that already had it. Now
`slice(0, 1)`, with a test pinning the head entry to the shipping version so
the next release cannot reintroduce it.

**A downgrade announced an update that never happened.** `changedVersion` was
true while the version filter matched nothing, emitting "updated to v0.20.0
(from v0.99.0)" over an empty list. Guarded on `notes.length > 0`; the marker
write stays unconditional.

**Stale `.role-<pid>` files accumulated forever.** `shutdown()` removes this
process's own role file, but it is only reached from stdin close, `SIGTERM`
and `SIGINT` - a `SIGKILL`, crash, OOM kill or power loss skips it, and
nothing swept the directory. My machine had six files for three live servers.
`server.ts` now sweeps role files whose pid answers no signal, before the
singleton lock is taken, so even a process that goes on to refuse and exit
tidies up on its way through. A live pid's file is never touched, and a name
that is not `.role-<number>` is left alone.

## Discovery

`review`, `manage` and `wizard --revoke` were reachable only by someone who
already knew their names — and the notice that would have told them was the
unreliable one above. `/whatsapp-channel:access` with no arguments now ends by
naming all three, and the skill description covers "add contacts", "add
groups", "bulk approve" so the phrasings people actually use trigger it.

## One thing to flag

The rename shipped without a CHANGELOG entry, so someone updating from 0.19.0
gets no word that the plugin renamed or that their old install stops receiving
updates. Since this PR is about exactly that failure, I wrote the 0.20.0 entry
for it — happy to drop that hunk if you would rather word it yourself.

## Checks

`bun test` — **298 pass, 13 skip, 0 fail**. Prettier clean on every file
touched; no new markdown fences, so no MD040. The statusline fix is verified
live on my machine (the old script printed nothing, the new one prints a green
`WA:primary`), and the new tests were each confirmed to fail against the
pre-fix source before being kept.

Not verified: that `systemMessage` renders in the terminal. The field is real
(size-capped separately in the CLI, counted separately in telemetry) and the
hook emits it correctly, but I have not watched a client display it.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "whatsapp-channel",
       "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.20.0",
+      "version": "0.21.0",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "whatsapp-channel",
   "displayName": "WhatsApp Channel for Claude Code",
   "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-channel:access.",
-  "version": "0.20.0",
+  "version": "0.21.0",
   "author": {
     "name": "Rich627"
   },
--- a/ACCESS.md
+++ b/ACCESS.md
@@ -108,6 +108,17 @@ Both fail with a clear error if the group's `roster` flag isn't granted.
 - Ctrl-C cancels cleanly at any point; nothing is written until every question on screen has been answered.
 - A group the wizard adds gets `requireMention: true` (only reply when addressed) — a more cautious default than `group add`'s own CLI default of `false` (reply to everything), deliberately: the wizard is the guided path for a less technical setup, the CLI is for someone already comfortable with explicit flags. Change it after the fact with `group add --mention` or `--no-mention`.
 
+### Taking access back (`wizard --revoke`)
+
+`bun scripts/access.ts wizard --revoke` is the same screen in reverse: it lists **everything currently configured** — every group in `access.json`'s `groups`, every contact in `allowFrom` — and you tick what should lose access. Unlike the grant screen it is never capped or ranked: a revoke list that left entries off could leave them permanently unrevokable.
+
+- **Leaving everything unticked removes nothing.** An empty selection is never read as "remove all".
+- Revoking a group keeps its `config.md` and `memory.md`, same as `group rm`, in case you add it back.
+- Revoking a contact also forgets their cached name and recency — unless another allowlist entry still resolves to the same person (someone allowlisted under both their `@lid` and phone form), in which case the cache is kept, because the surviving grant still needs it.
+- Ctrl-C cancels cleanly, and nothing is written until every question on screen has been answered.
+
+It reads the same two functions the in-session `/whatsapp-claude-channel:access manage` screen reads, so the terminal and the chat path cannot disagree about what is configured or what a revoke cleans up.
+
 Group/contact names and recency come from caches (`~/.whatsapp-channel/groups-meta.json`, `dm-activity.json`, `contacts.json`) that only the running server writes — automatically, as WhatsApp reports chat activity, no manual step needed once the account has connected at least once. If it's never connected yet, the wizard has nothing to show; pair it first. `groups-meta.json` is always written; `dm-activity.json` and `contacts.json` follow `WHATSAPP_CACHE_CONTACTS` (see "Names and privacy" above) — with it off (the default), the DM screen has nothing to rank until you turn it on.
 
 It's a terminal command, deliberately not a Claude Code skill: running it yourself, outside any chat, is what makes this true —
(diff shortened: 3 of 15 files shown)


Rich627:
Reviewed this end to end — it's in good shape and I'm happy to take it. What I verified, what I'd still like changed (one thing, non-blocking), and two answers you asked for.

## Verified

- **Tests:** ran the suite in a clean worktree off `219bcb2` — **311 pass, 0 fail**. (More than the 297 in your description, since the role-file sweep landed after you wrote it.) CI is green on all three checks.
- **Version pairing:** `marketplace.json` `plugins[0].version` and `plugin.json` both at `0.21.0`, base at `0.20.0`. This is the rule that has silently no-opped `plugin update` for users before, so I check it every time — correct here.
- **`availableVersion()` path derivation:** checked the relative walk against a real install on my machine. `cache/<marketplace>/<plugin>/<version>/` up four to `plugins/`, then `marketplaces/<name>/.claude-plugin/marketplace.json` — the depths line up exactly, and a repo checkout misses both paths and returns null rather than throwing. Deriving it from `import.meta.dir` instead of hardcoding was the right call.
- **Danger zone (`server.ts`):** both edits are contained. The stamp keeps line 1 byte-identical, so an un-updated statusline reader keeps working. `sweepStaleRoleFiles` only unlinks `.role-<integer>` whose pid answers no signal, with `pidAlive` treating EPERM as alive — so it cannot reap a live server's file it doesn't own. Running it before the lock is taken (so a refused server still tidies on its way through) is a nice touch and I'm glad it has a test pinning that ordering. Neither edit touches the connection lifecycle, the singleton lock, or access gating.
- **Access control:** the revoke path only ever takes access away, empty selection is a no-op, and `revokeCachedIdentity()` is line-for-line equivalent to what `remove` did — including the case that matters, one person in `allowFrom` under both their `@lid` and phone form, where revoking one form must keep the shared cache. Adding `wizard --revoke` to the set of subcommands the skill may not run is correct and I'd have asked for it if it weren't there.

## One thing I'd like changed

**A stamped role file outlives its server.** `roleFromOwnerStamp` matches on the owner pid and never checks the server pid in the filename. If a server is SIGKILLed and not restarted while its owning session stays alive, the file survives, the owner is still an ancestor, and the statusline keeps printing `WA:primary` for a server that isn't there. The old tree walk went blank in that case, which was at least honest.

The startup sweep doesn't cover it — that runs when a server *starts*, and here nothing starts. It should be cheap: `readRoleFiles` already has the filename, and `main()` already holds the process list, so dropping any file whose `.role-<pid>` isn't in `rows` closes it. Happy to take it as a follow-up if you'd rather land this as is.

## Your two questions

- **The 0.20.0 CHANGELOG entry:** keep it. The rename shipping without one is exactly the failure this PR is about, and your wording is accurate on the part that matters — an install under the old name keeps working but stops receiving updates. It'd be strange to ask you to fix the notice pipeline and then leave the notice empty.
- **`systemMessage` rendering:** fine unverified, and thanks for flagging it rather than rounding it up. Worst case it doesn't render and we're back to today's behaviour of nobody seeing the notice — no regression, and the "update available" notice repeating every session means we'd find out rather than it failing silently the way the old one did.

Three real bugs, each one found by using the thing rather than reading it, and the diagnosis on #3 (compound statusLine command means the shell is the wrapper's sibling, not its ancestor) is the kind of thing that stays broken for months. The intermittent-is-worse-than-permanent note on the wrapper matching is right, and stamping ownership instead of inferring it is the correct fix rather than a deeper tree heuristic. Appreciated.

PR #27, opened 2026-08-25
drop role files whose server pid is gone, since only a clean exit removes one

MrTig-afk:
The follow-up you asked for in #26: a stamped role file outliving its server.

## The bug

`roleFromOwnerStamp` matches on the owner pid on line 2 and never checks the
server pid in the filename. Kill a server without a graceful exit and never
restart it, and the file stays: the owning session is still alive, still an
ancestor, so the statusline keeps printing `WA:primary` for a server that
isn't there. The old tree walk went blank in that case, which was at least
honest.

The startup sweep can't cover it — that runs when a server *starts*, and here
nothing starts.

## The fix

Exactly where you pointed: `readRoleFiles` already has the filename and
`main()` already holds the process list, so it takes the live pid set and
drops any `.role-<pid>` that isn't in it. One `.filter`, no new process
scan. A non-numeric name isn't in the set either and drops out the same way —
this script only ever declines to believe a file, it never unlinks one.

## Tests

- New: `ignores a stamped file whose server is gone` — dead server pid, live
  owner stamp, expects silence. Verified against pre-fix source, where it
  fails with `WA:secondary`.
- Two existing e2e fixtures used dead server pids (`.role-4242`, `.role-4243`)
  and would have started passing for the wrong reason. Both now use live pids,
  so the stamp is still what decides them, not the new filter.

`bun test`: 299 pass, 13 skip, 0 fail. Prettier clean on every touched file.

## Version and changelog

0.21.1 in both manifests, plus the matching CHANGELOG entry — without a bump
`plugin update` never fetches this, which is the silent no-op you check for
every time.

That entry broke `first-ever run shows the NEWEST changelog entry, not the
oldest`, which asserted on `"--revoke"` — 0.21.0's text. Any new head entry
would break it, and the tempting "fix" is to paste the new text in and leave
the same trap for the next release. It now derives the head note from the
source, the way the version-pairing test right below it already does. Verified
it still catches the original bug by flipping `slice(0, 1)` back to
`slice(-1)`.

Happy to drop the version bump into a separate commit or leave the release to
you if you'd rather cut 0.21.1 yourself.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "whatsapp-channel",
       "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.21.0",
+      "version": "0.21.1",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "whatsapp-channel",
   "displayName": "WhatsApp Channel for Claude Code",
   "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-channel:access.",
-  "version": "0.21.0",
+  "version": "0.21.1",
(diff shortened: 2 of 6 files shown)


Rich627:
@codex review

PR #33, opened 2026-08-29
fix: transcribe voice/audio before the mention gate in requireMention groups

diogofrj:
## What & why

A voice/audio message carries no caption and WhatsApp gives it no native
`@`-mention slot, so `text` is empty by the time `gate()` runs. In a group
with `requireMention` on, `isMentioned()` tests that empty text and always
fails — the message is dropped before it is ever downloaded or transcribed,
no matter what is actually said in the recording (even the wake word
itself).

Text messages are unaffected. Only `requireMention` groups receiving
audio hit this — the same audio sent to a DM (no mention gate at all)
transcribes fine, which is what made the asymmetry visible: identical
voice note, DM works, group silently drops it. `diag.log` confirmed
`dropped inbound` entries at the exact timestamps of the group sends,
with nothing downloaded.

## Fix

In `handleMessage`, before the `gate()` call, detect the one shape where
the drop is otherwise certain (group, `requireMention` on, sender allowed
per `allowFrom`, no native mention already present) and transcribe once,
ahead of `gate()`, folding the transcript into `text` so `isMentioned()`
can see what was actually said instead of nothing. The existing eager
download/transcribe path further down (unchanged in its own logic) now
checks for that early result first and reuses it instead of downloading
and transcribing the same note twice.

`gate()` itself is untouched — same signature, same logic, same DM
handling. The new code only ever runs to decide a group-audio drop that
`requireMention` would otherwise make unconditionally, and only for live
messages (`!backlog`), matching the guard the normal media-download path
already uses. Allowlist/access gating invariant preserved: the early path
replicates the exact same `allowFrom` check `gate()` uses, so a
not-allowed sender's audio is never transcribed just to be dropped anyway.

## How I tested it

- `bun test` — 452 pass / 1 skip / 0 fail (full existing suite, unchanged).
- `bunx prettier --check` — clean (matches the file's existing style; the
  original file was already prettier-clean before my change, confirmed by
  diffing against a pristine `origin/main` copy).
- Manual trace-through against a real capture: reproduced the bug against a
  live instance first (group with `requireMention: true`, plain voice note
  → silently dropped, `diag.log` shows the drop at the same timestamp,
  nothing in the inbox), confirmed the code path (`gate()` runs on
  `extractText()`'s output, which is `""` for a bare `audioMessage`, before
  the download/transcribe block later in the function), then applied and
  reasoned through this fix against that same trace.
- Could not run `trunk check` — a fresh clone of this repo has no `.trunk/`
  config, so the launcher asks for `trunk init` rather than checking. I
  don't have access to whatever org-level config CI resolves, so please
  let CI's `trunk` job be the real check here; happy to adjust for
  anything it flags.
- Did not add an automated regression test: `gate()`/`handleMessage`
  aren't exported or exercised by the existing suite (nothing in
  `lib/*.test.ts` or `scripts/*.test.ts` touches them), and there's no
  existing harness for the live-socket-shaped parts of `handleMessage`.
  Open to pointers if there's a preferred way to cover this that I missed.

## Checklist

- [ ] I ran `trunk fmt` and `trunk check` and it passes. — see note above;
  could not run locally (no `.trunk/` config in a fresh clone), ran
  `bun test` + `prettier --check` instead.
- [x] Version bumped in BOTH files to the same new version (0.22.1 → 0.22.2, patch):
  - [x] `.claude-plugin/plugin.json` → `version`
  - [x] `.claude-plugin/marketplace.json` → `plugins[0].version`
- [x] Touched access/allowlist gating (see "Fix" section above for the
  invariant preserved) — did not touch connection lifecycle or the
  singleton lock.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -10,7 +10,7 @@
     {
       "name": "whatsapp-channel",
       "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.22.1",
+      "version": "0.22.2",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -2,7 +2,7 @@
   "name": "whatsapp-channel",
   "displayName": "WhatsApp Channel for Claude Code",
   "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-channel:access.",
-  "version": "0.22.1",
+  "version": "0.22.2",
   "author": {
     "name": "Rich627"
   },
--- a/server.ts
+++ b/server.ts
@@ -3572,6 +3572,61 @@ async function handleMessage(msg: WAMessage, backlog = false): Promise<void> {
   // because our passive lid-mapping.update listener missed it.
   await ensureLidResolved(senderJid);
 
+  // A voice/audio message carries no caption and no native WhatsApp
+  // @-mention — `text` is empty until it is transcribed. A mention-gated
+  // group would otherwise drop it on `text` alone before anyone finds out
+  // what was actually said, even when the recording opens with the wake
+  // word. Transcribe once, here, before the gate runs, so the mention check
+  // sees real content instead of nothing. Scoped tight to avoid wasted
+  // transcription: only when the drop is otherwise certain (group,
+  // requireMention on, no native mention already) and only for live
+  // messages, matching the `!backlog` guard the normal media path uses
+  // below.
+  let earlyAudioPath: string | undefined;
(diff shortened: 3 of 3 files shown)


Rich627:
@codex review

Rich627 (changes requested):
Request changes. The existing inline findings are valid merge blockers:

1. The pre-gate path runs before `gate()` checks `dmPolicy === "disabled"`, so it can download/transcribe group audio (including via an external provider) despite disabled mode dropping every message.
2. If the later mention gate rejects the transcript, the pre-gate raw audio file remains in `INBOX_DIR`; this contradicts `context: false`'s no-storage guarantee and leaks orphan files otherwise.
3. The 0.22.2 manifest bump lacks the matching head entry in `scripts/update-notice.ts`; `scripts/update-notice.test.ts` explicitly enforces that invariant, so the stated full test pass is not reproducible.

Please address these before merge.

Issue #1, opened 2026-04-07
Would a HOL skill validation workflow be useful for whatsapp-claude-plugin?


internet-dot:
If useful, I can open one PR that adds a validate-only HOL skill workflow.
It would only add one workflow file, keep publish mode off, and leave runtime code alone.
If that is not useful here, no problem.

m13v:
the permission relay concept (approving/denying tool requests from WhatsApp) is clever. biggest challenge with that flow is latency. Claude's tool calls have timeouts, and if the approval comes from WhatsApp even 30 seconds late, the session might have already timed out or moved on. we handle a similar pattern by batching tool requests and presenting them as a single approval message rather than one-by-one. a validation workflow makes sense since the WhatsApp input parsing needs to be bulletproof. a typo in a tool approval could accidentally grant permissions you didn't intend. I'd keep it lightweight though, just schema checks on the message format and a confirmation echo before executing.

m13v:
our WhatsApp MCP server that handles the bidirectional Claude/WhatsApp communication: https://github.com/m13v/whatsapp-mcp-macos

and the tool execution layer with the batched approval pattern I mentioned: https://github.com/m13v/fazm/blob/main/Desktop/Sources/Providers/ChatToolExecutor.swift

PR #34, opened 2026-09-02
Send documents with a real mimetype so Android WhatsApp can open them

dwsaxton:
Outbound attachments that aren't images or videos are sent as `application/octet-stream`, which Android WhatsApp shows as an unopenable "BIN" file.

This maps common extensions (pdf, md, txt, csv, json, html, zip, docx/xlsx/pptx, mp3/m4a/ogg/wav) to their mimetypes in the document branch of `reply`, and keeps `application/octet-stream` as the fallback for anything else, so behaviour for unknown types is unchanged.

Tested on 0.22.1 with a PDF sent to Android WhatsApp: shows and opens as a PDF.

### DIFF
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,8 +9,8 @@
   "plugins": [
     {
       "name": "whatsapp-channel",
-      "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
-      "version": "0.23.0",
+      "description": "WhatsApp channel for Claude Code \u2014 linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
+      "version": "0.23.1",
       "source": "./",
       "category": "channels"
     }
--- a/.claude-plugin/marketplace.json
+++ b/.claude-plugin/marketplace.json
@@ -9,7 +9,7 @@
   "plugins": [
     {
       "name": "whatsapp-channel",
-      "description": "WhatsApp channel for Claude Code \u2014 linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
+      "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control, pairing, allowlists, and group policy.",
       "version": "0.23.1",
       "source": "./",
       "category": "channels"
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,7 +1,7 @@
 {
   "name": "whatsapp-channel",
   "displayName": "WhatsApp Channel for Claude Code",
-  "description": "WhatsApp channel for Claude Code \u2014 linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-channel:access.",
+  "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-channel:access.",
   "version": "0.23.1",
   "author": {
     "name": "Rich627"
@@ -9,10 +9,5 @@
   "homepage": "https://github.com/Rich627/whatsapp-claude-plugin",
   "repository": "https://github.com/Rich627/whatsapp-claude-plugin",
   "license": "Apache-2.0",
-  "keywords": [
-    "whatsapp",
-    "messaging",
-    "channel",
-    "mcp"
-  ]
+  "keywords": ["whatsapp", "messaging", "channel", "mcp"]
 }
--- a/.claude-plugin/plugin.json
+++ b/.claude-plugin/plugin.json
@@ -1,13 +1,18 @@
 {
   "name": "whatsapp-channel",
   "displayName": "WhatsApp Channel for Claude Code",
-  "description": "WhatsApp channel for Claude Code — linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-channel:access.",
-  "version": "0.23.0",
+  "description": "WhatsApp channel for Claude Code \u2014 linked-device messaging bridge with built-in access control. Manage pairing, allowlists, and policy via /whatsapp-channel:access.",
+  "version": "0.23.1",
   "author": {
     "name": "Rich627"
   },
   "homepage": "https://github.com/Rich627/whatsapp-claude-plugin",
   "repository": "https://github.com/Rich627/whatsapp-claude-plugin",
   "license": "Apache-2.0",
-  "keywords": ["whatsapp", "messaging", "channel", "mcp"]
+  "keywords": [
+    "whatsapp",
+    "messaging",
+    "channel",
+    "mcp"
+  ]
 }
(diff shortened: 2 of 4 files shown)


Rich627:
@codex review

Rich627:
Thanks for this — the bug is real and the fix is in the right place. I checked every outbound send site and `server.ts:3151` is the only one that hardcoded `application/octet-stream` (the long-message overflow document at `:3103` already sets a real mimetype), so this closes the only gap. It also stays clear of the fragile parts of `server.ts` (connection lifecycle, singleton lock, access gating), and keeping `application/octet-stream` as the fallback means unknown types behave exactly as before.

Two things before I can merge:

**1. Version bump** (Codex flagged this too, and it's hard rule #1 in `AGENTS.md`)

`.claude-plugin/plugin.json` `.version` and `.claude-plugin/marketplace.json` `.plugins[0].version` are both still `0.23.0` — please bump both to `0.23.1`. Without it `plugin update` silently no-ops and nobody actually receives the fix. Note the CI `version-consistency` job only checks that the two files *match*, so it can't catch a missing bump — that one's on us as reviewers.

**2. Trunk check is red — it's prettier**

I diffed against `main` locally: `main` is clean, this branch has 4 formatting spots. `trunk fmt` (or `prettier --write server.ts`) fixes all of them:

- the extra blank line after `mimeToExt` — two consecutive blank lines get collapsed to one
- `.docx` / `.xlsx` / `.pptx` lines exceed the width; prettier wraps the value onto its own indented line

Everything below is optional — I'm happy to merge without any of it.

- **Consider putting `mimeForExt` in `lib/` with a test.** The repo keeps pure functions there with a colocated `*.test.ts` (`lib/sanitize.ts`, `lib/mentions.ts`, `lib/sent-log.ts`, …) and CI runs `bun test`. This is exactly that shape. (`AGENTS.md` still claims "No test suite" — that line is stale, my fault, I'll fix it.)
- **The two tables can drift.** `mimeToExt` (inbound/download) and `mimeForExt` (outbound) are inverses that already overlap on pdf/mp3/m4a/ogg. Deriving both directions from one table would be tidier, but that touches the existing download path, so I'd rather not widen this patch — just flagging it.
- `ext.toLowerCase()` is redundant: the caller already does `const ext = extname(f).toLowerCase()`. Harmless, just noting it.
- A few common extensions you could add while you're here: `.doc` / `.xls` / `.ppt`, `.xml`, `.yaml` / `.yml`, `.log`, `.tar` / `.gz`, and `.svg` (it isn't in `PHOTO_EXTS`, so it goes down the document branch).
- One question: `.mp3` / `.m4a` / `.ogg` / `.wav` now get `audio/*` but are still sent as a **document**, not an audio message. That's a strict improvement over `BIN` so I'm fine with it — but if the goal was in-app playback, the branch is what needs to change, not the mimetype. Your call; I don't think it belongs in this PR.

Push the version bump and `trunk fmt` and I'll merge. Thanks again for tracking this down and testing it on a real Android device.

Rich627:
Version bump is correct — `0.23.1` in both manifests and they match. Thanks.

Three small things left and then this is good to go:

**1. `trunk fmt` still hasn't run on `server.ts`.** All four spots from my last comment are unchanged — the extra blank line after `mimeToExt`, and the three long `.docx` / `.xlsx` / `.pptx` lines.

**2. The bump commit reformatted `.claude-plugin/plugin.json` and introduced a new prettier failure.** The `keywords` array got expanded onto four lines; prettier wants it back on one:

```json
  "keywords": ["whatsapp", "messaging", "channel", "mcp"]
```

Running `trunk fmt` (or `prettier --write server.ts .claude-plugin/*.json`) fixes 1 and 2 in one go.

**3. Cosmetic:** the em dash in the `description` of both JSON files got rewritten as a backslash-u escape sequence (`\u` followed by `2014`) instead of the literal character — looks like a JSON writer with ASCII escaping was involved. It's semantically identical (I diffed the parsed JSON of both files against `main`; the version string is the only real change), so nothing is broken, but these are the files we hand to the marketplace and I'd rather keep the diff to the version line. Please restore the literal character.

One heads-up on CI: the run for `efa0f670` is sitting at `action_required` because this is a fork PR, so the red X you may still see is from the first commit, not this one. I'll approve the run after your next push so we get a real signal.

None of this is on you conceptually — the actual fix has been right since the first commit. Once formatting is clean I'll merge.

dwsaxton:
Thanks for the careful review. Pushed b78433c: `prettier --write` on `server.ts` (blank line + the three long lines), and both manifests restored from `main` with only the version line changed (em dash literal again, `keywords` back on one line). Happy to move `mimeForExt` into `lib/` with a test in a follow-up if you'd like.

Rich627:
Formatting is clean now — Trunk check is green on b78433c, and both manifests diff against main on the version line only (em dash literal, `keywords` back on one line). Thanks for the quick turnaround.

One last thing, and it's the red X you're seeing. The Test job fails on exactly one test:

```
(fail) update-notice.ts > the newest CHANGELOG entry is the version actually shipping
 507 pass / 3 skip / 1 fail
```

The changelog here isn't a `CHANGELOG.md` — it's the hand-maintained `CHANGELOG` array at the top of `scripts/update-notice.ts` (~line 58), which drives the "what's new" notice shown at session start after an update. That test asserts its newest entry equals `plugin.json`'s version, precisely so a version bump can't ship without a note. Its head is still `0.23.0` while the manifests now say `0.23.1`, so it fails.

The fix is one entry at the top of the array:

```ts
const CHANGELOG: { version: string; notes: string[] }[] = [
  {
    version: "0.23.1",
    notes: [
      'Attachments that are not images or videos are now sent with their real file type. A PDF, .md, .csv, .zip or Office document previously arrived as an unopenable "BIN" file on Android WhatsApp.',
    ],
  },
  {
    version: "0.23.0",
    // ...
```

The notes are user-facing prose — what changed for someone using the channel, not the code path — so reword it however you like. `bun test scripts/update-notice.test.ts` confirms it, and `trunk fmt` afterwards.

That one's on me: I quoted hard rule #1 (bump both manifests) without mentioning the changelog pairing that the bump then trips. Sorry for the extra round trip.

On CI: "Plugin versions match" is still queued for fork approval, so it hasn't reported yet. It will pass — 0.23.1 matches in both files. I'll approve the run after your next push so all three checks finally report against the same commit, and then I'll merge.

dwsaxton:
Added the 0.23.1 CHANGELOG entry in scripts/update-notice.ts (bun test green locally), prettier run. Should be good for CI now.

Rich627:
LGTM. I approved the workflow run and all three checks are green on b0016cd — Test, Trunk check, and Plugin versions match.

The changelog note reads well too; naming the actual file types people send is better than what I sketched. Final diff is exactly the four files it should be: the mimetype map and its one call site in `server.ts`, the two manifests on the version line, and the new 0.23.1 entry.

Merging now. Thanks for finding this, for testing it on a real Android device, and for staying with it through three rounds of repo-specific rules that were nowhere in the contributing docs — that part was on us, not you.

The `lib/` + test follow-up would be welcome if you feel like it, along with any of the extra extensions I mentioned (`.doc`/`.xls`/`.ppt`, `.xml`, `.yaml`, `.svg`, `.tar`/`.gz`). No obligation either way.

PR #35, opened 2026-09-10
fix: make typing indicator opt-in (owner shown as typing in every chat)

abhinavjain-cmyk:
## Problem

`handleMessage` sends a `composing` presence update for every gated inbound message (server.ts, "Typing indicator" block). Group members see the owner's number as **"typing…" whenever anyone posts**, which reads as the owner actively watching every message. Several people in my groups noticed and asked about it.

## Fix

Gate the presence update behind a new optional `access.json` field `"typingIndicator": true`. Default is off, matching the least-surprising behavior for a silent-observer setup. Existing configs keep parsing (optional field, same pattern as `ackReaction`).

Versions bumped in both `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` (`plugins[0].version`) per AGENTS.md rule 1.

Invariant preserved (AGENTS.md rule 2): access gating untouched — the change only adds a read of an optional config field after the gate has already passed; connection lifecycle and singleton lock untouched.


Rich627:
@codex review

Rich627:
Thanks for the PR, and for the clear write-up. The `server.ts` change itself is right: it reads one optional field after the gate has passed and touches nothing else. Agreed that firing `composing` before Claude has decided anything is wrong for the silent-observer setup (`requireMention: false` groups, or allowlisted DMs).

A few things before this can merge:

**1. Flip the default to opt-out.** Every existing install would lose the typing indicator on update with the current default. Please make it `typingIndicator?: boolean` defaulting to **on**, disabled with `"typingIndicator": false` in `access.json`. Check it strictly (`access.typingIndicator !== false`) so a stray string in the file cannot change the behavior either way.

**2. CI is red on two jobs.**
- `Test`: `scripts/update-notice.test.ts` enforces that the newest `CHANGELOG` entry in `scripts/update-notice.ts` matches `plugin.json`'s version. Please add a `0.23.2` entry at the top describing this change (this is also what tells returning users the new switch exists).
- `Trunk check`: both `.claude-plugin/*.json` files were re-serialized: the `—` became the escape `\u2014`, `keywords` went multi-line, and the trailing newline was dropped. Please keep the diff to the two version lines only. b78433c in #34 is the same fix if you want a reference.

**3. Make it reachable without hand-editing JSON.**
- Add `typingIndicator` to `SET_KEYS` in `scripts/access.ts`. Note that `set` stores values as strings, so it needs explicit boolean parsing (`true` / `false`, reject anything else); otherwise `set typingIndicator false` would store `"false"`. A test in `scripts/access.test.ts` covering both values would be welcome.
- Document the key in `skills/access/SKILL.md` under `set <key> <value>`, and in the README one-liner that lists the other keys.

Happy to merge once those are in.

PR #36, opened 2026-09-15
Make the message-store cap configurable via WHATSAPP_MAX_STORE

triunfosalaroli-png:
## Summary
- `MAX_STORE` (the FIFO cap on the message key/proto store used by `download_attachment`, replies, and `message_id` lookups) was a fixed 500, shared across every chat.
- On a high-volume account (several active groups, hundreds of unread messages), that cap churns through in minutes, so `download_attachment` can fail with `Message not found in store` for media that hasn't actually expired on WhatsApp's side — just evicted locally by this store.
- Each stored entry is a message key plus a small proto, so a much larger cap is cheap. Made it configurable via `WHATSAPP_MAX_STORE`, default unchanged at 500 for backwards compatibility.
- Documented the limitation and the override in `USAGE.md`'s "Known limitations" section.

## Test plan
- [x] `bun server.ts` starts with no env var set, `MAX_STORE` still resolves to 500.
- [x] Ran with `WHATSAPP_MAX_STORE=5000` set and confirmed the store now holds more entries before evicting (verified in a local high-volume account).
- [ ] Maintainer call on whether the default itself should also move up.


server.ts
```diff
@@ -2227,7 +2227,14 @@ function ownerDisplayName(): string {
...
 
-const MAX_STORE = 500;
+// Shared across every chat, FIFO. A high-volume account (many active groups,
+// hundreds of unread messages) can churn through 500 entries in minutes, so a
+// message can hit "Message not found in store" on download_attachment well
+// before its media actually expires on WhatsApp's side. Override with
+// WHATSAPP_MAX_STORE if the default is too small for your traffic; each
+// entry is just a message key + a small proto, so a much larger cap costs
+// negligible memory.
+const MAX_STORE = Number(process.env.WHATSAPP_MAX_STORE) || 500;
```
Rich627: `Number(value) || 500` accepts values that violate the bounded-store invariant. For example, `WHATSAPP_MAX_STORE=-1` makes every newly stored message get evicted immediately, while `WHATSAPP_MAX_STORE=Infinity` disables eviction and allows unbounded memory growth; fractional values also produce surprising effective limits. Please parse this once and accept only a finite positive integer (for example with `Number.isSafeInteger(value) && value > 0`), otherwise fall back to 500 or fail with a clear startup error.

Rich627 (changes requested):
Thanks for addressing the high-volume eviction issue. The current CI run passes, but there are two blocking issues:

1. **Release version was not bumped.** Both `.claude-plugin/plugin.json` and `.claude-plugin/marketplace.json` remain at `0.23.1`, the same as the base commit. The CI check only confirms that the two values match each other; it does not check that the plugin version is newer than the base. As a result, existing users will not receive this feature through `plugin update`. Since this is a feature, please bump both plugin versions together (normally to `0.24.0`).

2. **`WHATSAPP_MAX_STORE` needs validation.** See the inline comment. Please accept only a finite positive integer and add focused coverage for unset, valid, negative, non-numeric, fractional, and `Infinity` values.

CI observed: Plugin versions match, Test, and Trunk check all passed on run 34922176205.

Rich627 (approved):
Thanks for addressing the feedback so quickly and for the thorough test coverage! Really appreciate your contribution to handling high-volume accounts. Everything looks great!

PR #37, opened 2026-09-15
catch_up: expose message_id for messages with an attachment

triunfosalaroli-png:
## Summary
- `catch_up` builds its per-message line with its own inline formatter, which never includes the message id — even for entries with `attachment_kind` set.
- `unreplied` already exposes it correctly (`formatMessages` appends `message_id=...`), so this is just an inconsistency between the two, not a design choice.
- Effect: `download_attachment` needs that id as `file_id`. Once a message with media drops out of the unreplied set — the owner replies to it by hand on their phone, or it just ages out — `catch_up` is the only tool left that can surface it, and it comes back with no id to act on. The agent can see "(voice attachment)" happened but has nothing to pass to `download_attachment`.
- Real trigger: this is exactly what happened to me — an owner's hand-reply on WhatsApp cleared a voice note from `unreplied` before I could look at it in the same turn; `catch_up` afterward showed the voice note happened but gave me no id to fetch it with.
- Fix: reuse the same `message_id=${e.id}` suffix `formatMessages` already uses, on the attachment branch of `catch_up`'s line formatter. One-line, mechanical, matches the existing pattern exactly.

## Test plan
- [x] Traced both formatters (`formatMessages` used by `unreplied`, and `catch_up`'s inline one) side by side to confirm the id is only missing from `catch_up`'s path, not a MessageLogEntry gap.
- [ ] Did not spin up a second live WhatsApp pairing in this fork to re-verify end-to-end — the change is a one-line string-template addition matching an existing, already-working pattern in the same file, and I didn't want to risk a second device link against a real account for this. Happy to do that if the maintainer wants it before merging.


Rich627:
@codex review

Rich627:
Thanks for the fix and the clear explanation! I've bumped the plugin version to `0.23.2` and added the CHANGELOG entry. Ready to merge once CI finishes.

PR #39, opened 2026-09-17
fix(markdown): run italic before bold so bold can render

### LINKED ISSUE #38: markdownToWhatsApp: italic rule consumes bold rule's output, so bold never renders
## Summary

`markdownToWhatsApp()` can never produce bold output. The italic rule runs
immediately after the bold rule and re-consumes the bold rule's own output, so
every `**bold**` span arrives as `_italic_` instead.

Present in `0.22.1`, `0.23.1` and `0.24.0` — `server.ts` lines 2095-2101 on
current `main`.

## Cause

```js
// Bold: **text** or __text__ → *text*
result = result.replace(/\*\*(.+?)\*\*/g, "*$1*");
result = result.replace(/__(.+?)__/g, "*$1*");

// Italic: *text* (single) or _text_ → _text_
// Only match single * not preceded/followed by * (to avoid conflicts with bold)
result = result.replace(/(?<!\*)\*(?!\*)(.+?)(?<!\*)\*(?!\*)/g, "_$1_");
```

The bold rule rewrites `**text**` to WhatsApp's `*text*`. The italic rule then
runs on that result. Its guards (`(?<!\*)` / `(?!\*)`) were written to avoid
matching the *input* `**text**`, but by this point the doubled asterisks are
gone — what remains is a clean single-asterisk pair, which is exactly what the
italic rule matches.

The comment "to avoid conflicts with bold" describes the intent, but the guards
only protect against un-converted input, not against the bold rule's output.

## Reproduction

```
input:  **hello**
step 1: *hello*    (bold rule)
step 2: _hello_    (italic rule — undoes step 1)
sent:   _hello_    → renders italic in WhatsApp
```

No input produces bold. I tried single, double and triple asterisks, `__bold__`,
and zero-width-space padding; all resolve to `_text_` or to mangled output.

## Suggested fix

Run the italic rule *before* the bold rules. A `**bold**` span cannot match the
italic rule — its asterisks are adjacent, so both lookarounds fail — so italic-
first leaves bold input untouched, and the bold rule then converts it.

```js
  // Italic: *text* (single) or _text_ → _text_
  // Runs BEFORE bold: a **bold** span can't match this (its asterisks are
  // adjacent), while the bold rule's *output* would.
  result = result.replace(/(?<!\*)\*(?!\*)(.+?)(?<!\*)\*(?!\*)/g, "_$1_");

  // Bold: **text** or __text__ → *text*
  result = result.replace(/\*\*(.+?)\*\*/g, "*$1*");
  result = result.replace(/__(.+?)__/g, "*$1*");
```

Verified against the current rules, ported 1:1:

| input | current | italic-first |
|---|---|---|
| `**bold**` | `_bold_` ❌ | `*bold*` ✅ |
| `*italic*` | `_italic_` ✅ | `_italic_` ✅ |
| `**a** and *b*` | `_a_ and _b_` ❌ | `*a* and _b_` ✅ |
| `plain` | `plain` ✅ | `plain` ✅ |

Applied locally and confirmed end-to-end: bold now renders as bold in WhatsApp,
and a single-asterisk span in the same message still renders italic.

## Related

The `// Headers → bold` rule above emits `*$1*`, so headings hit the same
problem and render italic. Changing it to `**$1**` would make it bold as
labelled once the reorder is in — I left it alone locally since I only needed
inline bold, but it looks like the same root cause.

Happy to open a PR if useful.

tomerikko:
Fixes #38.

`markdownToWhatsApp()` could never produce bold: the bold rule rewrites `**text**` into WhatsApp's `*text*`, and the italic rule ran immediately afterwards on that output, matching it just as readily as a genuine italic span. Every `**bold**` arrived as `_italic_`.

Reordering fixes it **without touching either regex** — a `**bold**` span cannot match the italic rule, because its asterisks are adjacent and both lookarounds fail. So italic-first leaves bold input untouched, and the bold rule then converts it.

| input | before | after |
|---|---|---|
| `**bold**` | `_bold_` ❌ | `*bold*` ✅ |
| `*italic*` | `_italic_` ✅ | `_italic_` ✅ |
| `**a** and *b*` | `_a_ and _b_` ❌ | `*a* and _b_` ✅ |
| `plain` | `plain` ✅ | `plain` ✅ |

**Invariant preserved:** single-asterisk and underscore italics convert exactly as before. Only the order of two independent replacements changes; neither pattern is modified.

Verified end-to-end on a live deployment — bold renders as bold in WhatsApp, and a single-asterisk span in the same message still renders italic.

Versions bumped per AGENTS.md rule 1: `marketplace.json` `plugins[0].version` and `plugin.json` both 0.24.0 → 0.24.1 (patch, bug fix).

Note: the `// Headers → bold` rule just above emits `*$1*`, so headings hit the same problem and render italic. Left alone here to keep this change minimal — happy to include it if you'd like.


Rich627:
@codex review

Rich627:
Thanks for the fix, @tomerikko — nice catch on the italic rule re-matching and undoing bold spans. Merging now.

PR #40, opened 2026-09-17
Quiet intake: counts at session start, one chat on request, one 7-day horizon (0.25.0)

MrTig-afk:
## The problem

At session start the channel read out every waiting message in every chat. With a busy group on
the allowlist that is a wall of text before the user has said anything, it leaks the contents of
private chats into a transcript that was only asked "anything new?", and an unanswered message
expired after 24 hours - so the one thing the user had not got to yet was the first thing to
disappear.

## What changed

**Session start is counts only.** `catch_up` with no argument prints one line per chat that has
something waiting - display name and count, an `@` where a mention-gated group actually addressed
you - plus the open items from `tasks.md`. No message text, no chat_id. A trailing line says how
many chats had something waiting that has since aged out, so a long absence does not read as
silence.

**One chat on request.** `catch_up chat=<name>` shows that chat's recent two-way conversation:
the last N messages from each side (default 5, `limit` up to 100), merged in time order, with
messages still waiting for a reply taking the inbound slots first so what the count was about is
what you see. A name that matches several chats returns the candidates (a DM by its number masked
to the last four digits, a group by its id) and asks, rather than guessing; the masked handle and a
clipped counts-row name are both accepted back. A renamed chat is listed under its current name.

**One 7-day horizon.** The separate 24-hour expiry for unanswered messages is gone; every stored
line lives 7 days. `WHATSAPP_MESSAGE_TTL_DAYS` (1-30, clamped, logged) changes that, and the same
horizon prunes `inbox/`. Replies you type on your phone stay readable in full for as long as the
line is kept instead of fading after an hour. `~/.whatsapp-channel/.aged-out-chats.json` remembers
- as a truncated hash, never the id - which chats lost a waiting message to the prune, for 30 days.

**`wait_for_messages` waits for new arrivals.** A caller is one connection: the first call returns
whatever is already unreplied (a fresh terminal is never starved), later calls return only what
arrived since, and the identity goes with the connection so a reconnecting terminal sees the
backlog again. Before this it returned the same backlog instantly on every call for as long as
anything was unanswered.

**Media without a caption reads as `[photo]` / `[voice]` / `[video]` / `[file]`** in `catch_up`
and `unreplied`, on every path (eager download, backlog, download failure), not as an empty line.

**Update banner.** One short line per change with a blank line between bullets; every existing
entry rewritten to that shape (0.22.0 previously stacked thirteen paragraphs). The 0.23.0 note that
described a permissions block as fixed while its tool ids still matched nothing is corrected, and
the ids in USAGE.md now match what Claude Code registers (`mcp__plugin_whatsapp-channel_whatsapp__*`).

**First run points at the access screen.** Session-start onboarding names the terminal wizard as
the normal route; a session with no terminal to open (Remote Control, headless, piped) prints the
exact command instead of hanging.

**`unreplied` / `wait_for_messages` are capped at the newest 100 per call**, after any `chat_id`
filter, and say how many more are waiting.

### Access and approvals (please read these closely)

- `access set owner` accepts only a contact already on the allowlist and stores the allowlist's own
  spelling. A mistyped digit is a well-formed jid, and before this it pointed every permission
  request - up to 500 characters of the command being approved - at a stranger, indefinitely.
- The stored owner is revalidated on every read. Removing that contact from the allowlist used to
  leave requests going to them while their replies were dropped at the gate, so the tool call hung
  with no diagnostic. Now requests go to the linked account (your own note-to-self, where both
  approval routes already work); before the connection opens they are not sent at all and the
  reason is logged. `access status` flags a stored owner that has left the allowlist.
- The access gate and the CLI compare jids the same way (Baileys' own normalisation on both sides,
  an unparseable jid matches nothing), so a value the CLI accepted is a value the server honours.
  This does not widen who is allowed: two spellings of the same account compare equal, and no jid
  of a different account matches an entry it did not match before.
- A group subject containing a line break can no longer forge an extra row in the counts view or an
  extra entry in `unreplied`; lines written before 0.23.0 were never sanitised and are still on disk.

### Cron

- `*/N` starts at the field's minimum, so `*/2` in day-of-month is the 1st, 3rd, 5th and `*/3` in
  month is Jan/Apr/Jul/Oct, as in crontab. `9/2` fires at 9, 11, 13 - the base used to be parsed,
  validated and discarded.
- The `(cron: "expr")` form the config.md header has always documented is now actually parsed
  (it was silently dropped), wins over prose that looks like a schedule, and accepts `9-17/2` and
  day-of-week 7. A bullet that yields no schedule is reported by `doctor`, not dropped.

### Tooling

- `doctor` can no longer print nothing at all; its inbox warning no longer claims `inbox/` is never
  pruned (a running primary prunes it hourly; the 500 MB threshold assumes the default horizon).
- `patch-baileys.mjs` matches targets whitespace-tolerantly and exits non-zero only for a target
  that is genuinely absent - previously a missing file was skipped silently and postinstall still
  succeeded.
- The watchdog and whisper reference scripts are tightened; both are copied out of the repo by
  users, so a running deployment does not pick them up until re-copied.

### Merged with main at 0.24.1

`main` released 0.23.2, 0.24.0 and 0.24.1 while this branch was in review, so this PR is
merged with it (#36, #37, #39 all apply unchanged) and the release is **0.25.0**, above what
users already have. Two things in that merge are mine:

- The three new banner entries are split into one-line notes, the shape every other entry
  on `main` now has, with the same content.
- `markdownToWhatsApp` had two defects that predate #39 and reproduce identically on either
  side of it: the header rule emitted `*Title*` before the italic rule rewrote it to
  `_Title_`, so every heading arrived italic; and the italic rule had no flanking check, so
  `a * b * c` became `a _ b _ c`. The header rule now runs after the italic rule, and the
  italic rule refuses whitespace next to either asterisk. Checked by running the real
  function source over ten cases (headings, bold inside a heading, `__bold__`, `*italic*`,
  spaced asterisks, strikethrough, inline code, plain text).

## Tests

`bun test`: 645 pass / 14 skip / 1 fail across 660, at 2e692c4 on the merged tree. The
one failure is `doctor.test.ts` "notify hook present but not executable → WARN with chmod fix",
which cannot be represented on a filesystem with no executable bit and fails identically on
untouched `origin/main` on this Windows machine; it is not touched by this branch. Type-checked
with `tsc --strict` (the only errors are `import.meta.dir` / `Bun` without bun-types installed,
pre-existing).

## Not in this PR

AI-agnostic packaging for Codex / Gemini CLI / Windsurf / Cursor (the `!` prefix printed by
onboarding is Claude-Code-only and is deliberately not printed). Twelve defects found on
`origin/main` during review are drafted as fork issues and not fixed here except where this branch
touched the same lines (cron step base, patch-baileys silent skip, `set owner` accepting a group
jid, and the two `markdownToWhatsApp` defects above).


Rich627:
Thanks for this — the quiet-intake behavior is a nice improvement. A few concerns from review, roughly in order of severity:

**1. Revoked access can still approve pending requests via emoji reaction (`server.ts:5450`)** — MEDIUM
The text-reply approval path re-checks `access.allowFrom` before honoring an approval, but the emoji-reaction approval handler doesn't perform the same re-check. If someone is removed from the allowlist while they still have a pending "approve" reaction available (or a stale request), they can approve new permission requests through the reaction path even though they no longer have access. This is effectively the same door the PR closes for text replies, left open for reactions.

**2. Permission requests can still be silently dropped (`server.ts:3182-3205`)** — MEDIUM
The new fail-closed diagnostic only checks for `!owner`. If `owner` is truthy but `sock` is falsy (e.g. right after opening a second terminal, or during the brief window before the connection is fully established), the permission request is dropped with no log and no notification — the same class of "message vanishes silently" bug this PR is meant to fix, just relocated rather than eliminated.

**3. CLI and server disagree on allowlist membership for bare-number JIDs** — LOW/MEDIUM
`scripts/access.ts` and the server's own check (around `server.ts:1475-1489`) normalize bare-number JIDs differently. This can make `access allow` / `access status` report a number as allowed when the server's actual runtime check disagrees, or vice versa — confusing when auditing who currently has access.

**4. Case-sensitivity bug in chat name lookup (`lib/message-view.ts`, `nameMatches`/`resolveChat`)** — MEDIUM
Looking up a chat by name for the on-request detail view is case-sensitive, so a chat name typed with different casing than stored won't resolve.

**5. Smaller/non-functional nits:**
- `lib/message-view.ts`'s `oneLine` doesn't strip control characters, which can distort console/log output.
- The ambiguity-warning row limit could use a second look.
- A doc comment on `countAgedOut` appears stale relative to current behavior.
- `takeUnseen` ordering is worth double-checking against intended semantics.

Happy to help verify fixes for #1 and #2 in particular, since those are the two that touch access control directly.

MrTig-afk:
Thanks for the careful review. All of it is addressed in 0.25.1 (9f4a3dd, 2058b0e):

**1. Reaction approvals after revocation.** Confirmed: typed replies only reach `claimPermission` after `gate()`, reactions never pass it. The reaction handler now re-checks `allowFrom` itself. The one exception is `fromMe` (the linked account), which the typed path also trusts, so the note-to-self flow keeps working.

**2. Silent drop with no socket.** Confirmed. That case is now logged, with a hint to answer in the terminal. I didn't add relaying permission requests from a secondary terminal to the primary: that is a new IPC message plus routing the answer back, and felt like its own PR. Happy to do it if you want it.

**3. CLI vs server on bare numbers.** Confirmed: `normalizeJid` passes a bare number through, while `isAllowedJid` fails closed on it. I kept the server strict and fixed the CLI: `access allow` refuses a bare number and suggests the `@s.whatsapp.net` form, and `status` flags any bare entry already in `access.json` as matching nobody. A bare `owner` is no longer shown as allowlisted.

**4. Case-sensitive chat lookup.** The one current caller lowercases first, so it wasn't reachable from `catch_up`, but `resolveChat` now lowercases its own input so the contract doesn't depend on the caller.

**5. Nits.**
- `oneLine` now strips control characters (`\p{Cc}`: ESC, NUL, DEL, C1) as well as whitespace. Emoji joiners are kept. Test added.
- The ambiguity limit (8) stays. Its comment claimed rows carry raw jids, but they have been masked for a while, so the comment is corrected.
- `countAgedOut`'s doc was stale, and so was `updateAgedOut`'s. Both now describe the "handled" early-clear rule the code actually has.
- `takeUnseen` ordering checked: `byTs` compares `toISOString` stamps (fixed format, so lexicographic order is chronological), the sort is stable, and the newest `limit` are handed. No change.

While in there I also fixed three pre-existing issues on `main`: sender-set `attachment_mime` / `reply_to_id` / `reply_to_sender` now go through `safeName` before the `<channel>` envelope; `startIpcListener` waits for the bind to settle, so a failed bind is reported as down; and `logOwnerHandReply` marks the chat replied before its group-name await, not after.

Tests: 648 pass locally (the one failure is a Windows-only exec-bit case in doctor.test.ts that fails on `main` too). I'd appreciate a second look at #1 and #2 as you offered.

Rich627:
Thanks for this — and for how thorough the round-2 write-up was.

I re-checked the two access-control items at source: the reaction re-check is
right (`reaction.key` is the server-stamped key, so trusting `fromMe` there is
sound), and the no-socket case is loud now instead of silent. Normalisation in
`isAllowedJid` doesn't widen across accounts, and the
`mcp__plugin_whatsapp-channel_whatsapp__*` ids are correct — the old ones
matched nothing.

Merging. Two follow-ups, neither blocking:

- `settings.json` only allowlists 5 tools, and `catch_up` is now the
  session-start intake, so it prompts every time. I'll open an issue.
- Yes please on the permission-request IPC relay as its own PR, if you're
  still up for it.

Thanks again for the contribution.

